PR #24, opened 2014-08-11
Add margin option.

phallguy:
Instead of adjusting the size of individual sprites, the margin option simply increases the y/x offsets between each sprite, retaining the original image size.


phallguy:
Thanks for the hard work. I've just read the docs but so far this looks great.

PR #49, opened 2015-01-29
Adds `display` property to generated css files

megapctr:
This should make sprites usable out of the box, without additional css tinkering.

### DIFF
--- a/tmpl/css/sprite.css
+++ b/tmpl/css/sprite.css
@@ -1,10 +1,12 @@
 {{#common}}.{{mixin}} {
+	display: inline-block;
 	background-repeat: no-repeat;
 	background-image: url({{{sprite}}});
 }
 
 {{/common}}{{#shapes}}{{#selector.shape}}{{expression}}{{^last}},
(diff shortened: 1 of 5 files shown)


megapctr:
When you'd try to use something like `<span class="svg-my_icon svg-my_icoon--dims"></span>`, the icon won't be visible. Adding `display: block` property would get it to show, but then the icon wouldn't align nicely with text.

jkphl:
Well, but it totally depends on your individual use case which element(s) and which display type(s) you prefer. Imagine someone using `<div>`s instead of `<span>`s, intentionally, having block display by default (and this is not a rare case!). He wouldn't want to have `display: inline-block` set on his icons by default.

My suggestion: Use a common CSS class for all your icons (e.g. `icon`) and have **one central CSS rule** in your custom stylesheets for setting `display: inline-block`. Maybe you could even do without an additional class (e.g. `span[class^=svg]`).

Please understand that I won't accept this pull request as the proposed modification is too specific and would limit other users.

PR #67, opened 2015-04-10
SVGO Version Bump

Meesayen:
The latest `svgo` patch fixes some issues with svg paths optimization. Much needed, otherwise resulting in glitched svg sprites.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
 		"xmldom": "^0.1.19",
 		"xpath": "^0.0.9",
 		"vinyl": "^0.4.6",
-		"svgo": "0.5.0",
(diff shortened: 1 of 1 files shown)


jkphl:
@Meesayen Thanks for reporting this. I already noticed the new SVGO release. However, as the Travis build failed, I will need to investigate this further before pushing out a new release. I'll try to do so tonight.

Cheers,
Joschi

jkphl:
The new SVGO build still disrupts some of the example SVG files included in the package. That's why I had restricted the SVGO dependency to 0.5.0. Disabling all the SVGO plugins doesn't solve this problem. I'll probably have to investigate this further — or even file an SVGO bug.

PR #72, opened 2015-04-16
Fix issue with view box negative values

rleite:
The current viewbox parsing dose not alow negative values. This breaks the sprite when calculating padding.

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -318,7 +318,7 @@ SVGShape.prototype._initSVG = function() {
 	var viewBox				= this.dom.documentElement.getAttribute('viewBox');
 	if (viewBox.length) {
 		// shouldn't this be .split(/ +/)
-		viewBox				= viewBox.split(/[-^\d\.]+/);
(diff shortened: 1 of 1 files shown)


jkphl:
While your point is very valid as a whole — *svg-sprite* should accept negative `viewBox` values — the regular expression that you provide is wrong. Should be `/[^-\d\.]+/` instead. I'll fix this myself, however. Thanks for reporting!

lib/svg-sprite/shape.js
```diff
@@ -317,7 +317,8 @@ SVGShape.prototype._initSVG = function() {
 	// Determine the viewbox
 	var viewBox				= this.dom.documentElement.getAttribute('viewBox');
 	if (viewBox.length) {
-		viewBox				= viewBox.split(/[^\d\.]+/);
+		// shouldn't this be .split(/ +/)
+		viewBox				= viewBox.split(/[^-\d\.]+/);
```
rleite: Here i did change it right.

rleite:
Thanks anyway!

Issue #74, opened 2015-04-20
Defs block for extracting gradients.


jaywolters:
With Symbols is it possible to configure svg-sprite to extract all the gradients &lt;linearGradient> (etc) into a &lt;defs> block -- or will I need to use Cheerio for that?

jkphl:
Hey @jaywolters,

it's not really within *svg-sprite*'s scope to perform SVG manipulations like that. You could, however, use a custom `transform` to operate on each single SVG shape before it's compiled into the sprite. See [here](https://gist.github.com/jarcoal/0f8d205598401baaf414) for a recent example of how to re-color a shape (used with [grunt-svg-sprite](https://github.com/jkphl/grunt-svg-sprite), but it'll work with *svg-sprite* just the same way).

*svg-sprite* doesn't use Cheerio but operates on a raw DOM implementation instead ([xmldom](https://github.com/jindw/xmldom)). Please see [here](https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md#custom-callback-transformation-function-values) for some info about custom callback transformations.

Finally, I am considering adding a [global post-transformation](https://github.com/jkphl/svg-sprite/issues/64) on the final sprite, but that has to be done yet.

Hope this helps!?

Cheers,
Joschi

jaywolters:
Using the Symbol method and when working with pre-colored Icons that use gradients-- moving the gradients to a &lt;Defs> block is essential for FF and IE to render the SVG correctly. This problem has been brought up in a few SO posts and grunt/gulp-svgstore made the change just recently. Thanks for your response.

jkphl:
Good to know, thanks for this info. You don't happen to have some resources handy elaborating on the topic? I might consider hardwiring it then ...

jaywolters:
Refer to David Bushell's <a href="http://dbushell.com/2015/01/30/use-svg-part-2/ ">blog post</a> on SVG and MDN <a href="https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Gradients">using gradients in SVG</a> you will see where it shows the SVG markup and how all the Gradients are placed inside of the &lt;defs> block. <br />
[...]

PR #81, opened 2015-05-07
add a new mode that combines the view and the symbol mode.

flobacher:
through the use of 'use'-tags, filesize increases only very little

### DIFF
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -17,43 +17,43 @@ var _									= require('lodash'),
 	winston								= require('winston'),
 	/**
 	 * Sprite types
-	 * 
-	 * @type {Array} 
+	 *
+	 * @type {Array}
 	 */
-	spriteTypes							= ['css', 'view', 'defs', 'symbol', 'stack'],
+	spriteTypes							= ['css', 'view', 'defs', 'symbol', 'stack', 'symboluse'],
 	/**
 	 * Default list of transformations
-	 * 
+	 *
 	 * @type {Array}
 	 */
 	defaultTransform					= ['svgo'],
 	/**
 	 * Default SVG configuration
-	 * 
+	 *
 	 * @type {Object}
 	 */
 	defaultSVGConfig					= {
 		/**
 		 * Add a DOCTYPE declaration to SVG documents
-		 * 
+		 *
 		 * @type {Boolean
 		 */
 		doctypeDeclaration				: true,
 		/**
 		 * Add an XML declaration to SVG documents
-		 * 
+		 *
 		 * @type {Boolean}
 		 */
 		xmlDeclaration					: true,
 		/**
 		 * Namespace IDs in SVG documents to avoid ID clashes
-		 * 
+		 *
 		 * @type {Boolean}
 		 */
 		namespaceIDs					: true,
 		/**
 		 * Add width and height attributes to the sprite SVG
-		 * 
+		 *
 		 * @type {Boolean}
 		 */
 		dimensionAttributes				: true
@@ -61,7 +61,7 @@ var _									= require('lodash'),
 
 /**
  * SVGSpriter configuration
- * 
+ *
  * @param {Object} config				Configuration
  */
 function SVGSpriterConfig(config) {
@@ -92,15 +92,15 @@ function SVGSpriterConfig(config) {
 			})]
 		});
 	}
-	
+
 	this.log.debug('Started logging');
 
 	this.dest							= path.resolve(config.dest || '.');
-	
+
 	this.log.debug('Prepared general options');
-	
+
 	this.shape							= ('shape' in config) ? _.assign({}, config.shape || {}) : {};
-	
+
 	// Parse meta data (if configured)
 	if (('meta' in this.shape) && !_.isPlainObject(this.shape.meta)) {
 		var meta						= _.isString(this.shape.meta) ? path.resolve(this.shape.meta) : null,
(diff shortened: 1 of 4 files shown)


flobacher:
maybe the name of the new mode should be changed to combined, or something else, but the idea is to be able to reference symbols via use tags in the html and to be able to use the same spritesheet via the help of the generated positions (view mode) in css for e.g. in an ::after selector.
i am really looking forward to hear your feedback.

jkphl:
Hi @flobacher,

thanks for these suggestions. Due to my current workload (and some travels the next days) I'll probably not be able to review this before mid of next week. Thanks for your patience!

Cheers, Joschi

flobacher:
Hi Joschi,

no need to hurry. just think such a combined spritesheet is realy handy.
u can reference symbols via their id in use tags.
and in css you can use the background-positions calculated from the viewmode.
and since it is all combined in one file, it will only be downloaded once.
this way it is also possible to generate a png fallback from the spritesheet,
which was not possible via the symbol mode alone.

hope you like it!

Cheers, Flo

Zitat von Joschi Kuphal <notifications@github.com>:

flobacher:
I just renamed the mode to universal, because its universal nature (symbols, views and positions all in one spritesheet) reminded me of the universal module definition pattern for js modules. I hope you think as well, that this is a better name for this mode.
cheers, flo

jkphl:
Hi @flobacher,

first of all sorry for not being more responsive! It really took me until now to find the time to thoroughly review your suggestions.

Although there are some parts that I would solve a little different, I really like your approach. In fact I wonder if we couldn't even take this a little further. Let me explain:

1. The **css** mode is equivalent to your `<symbol>`s in combination with the `<use>` elements..
2. The **view** mode is an extension to the **css** mode adding the `<view>` elements.
3. The **defs**, **symbol** and **stack** modes are mutually exclusive, but could probably all be extended by the `<use>` plus `<view>` features (not sure about the **stack** mode, but I'll try).

So why not generally configure a sprite by combining "flags" like this (hope you understand the EBNF like try of an explanation):

```
Sprite                    = UniversalSprite | BackgroundSprite ;
UniversalSprite           = LibrarySprite , { LibraryBackgroundSprite } ;
LibrarySprite             = DefsSprite | SymbolSprite | StackSprite ;
BackgroundSprite          = CssSprite, { ViewSprite } ;
LibraryBackgroundSprite   = UseLibrarySprite, { ViewSprite } ;
```

In words:

1. First you have to decide wether you want one of the "library based" sprites (**symbol**, **defs** or **stack**) or just a -"simple" CSS sprite for use as background image (**css** / **view**).
2. In case of the latter you have to decide whether you want the `<view>` elements or not (IMHO the only reason against them could be to save some bytes ...).
3. If you go for the library based sprite, you may additionally enable the background capabilities (including or excluding the `<view>` elements).

This could elmininate the need for the different modes altogether and make the configuration more like a flag based process. Maybe that's a lot simpler than the current way of configuring. What do you think? Looking forward to your thoughts!

Cheers,
Joschi

flobacher:
Hi @jkphl, 
thanks for taking the time to get back on me.. I am sure, you can do a much better way of integrating that new mode into your code than me! And yes, I like your flag based approach of combining the diffrent modes into a single file very much! Just one comment.. I think at Point 3 of your explanation I would just add for completeness, to add those bg capabilities via <use > tags inside the spritesheet, so that the shapes don't have to be duplicated.. and the <use>, just like the <view> just adds a couple of bytes, so that should not hurt anyone.. I would really love to see that in svg-sprite =)

jkphl:
Hey @flobacher,
just to let you know: I'm planning a major overhaul of *svg-sprite* somewhere around August, mostly focusing on usability and ease of configuration. I'll try to combine all this then.
Cheers,
Joschi

flobacher:
Great!!!

Let me know, if you need any help!

Cheers,
Flo

Zitat von Joschi Kuphal <notifications@github.com>:

stikoo:
This is exactly the functionality I'm after. I'm currently moving to a symbol workflow but would like to retain the ability to use the CSS method as well. Will just have to have 2 output modes for now :(

XhmikosR:
@flobacher @Kreeg I rebased this and updated to match the current code, but, unfortunately, I can't push to @flobacher's master branch: https://github.com/svg-sprite/svg-sprite/compare/pr/81

We still need tests, though.

Kreeg:
@XhmikosR do you mean screenshot tests?

XhmikosR:
Whatever tests are needed, not sure exactly. I just don't see any tests is all :)

If you plan to land it later @Kreeg I guess we should close this PR and you use the branch I pushed to our repo.

Kreeg:
@XhmikosR ok i will land it later. And I will close this PR then. For now I think it should be opened

Kreeg:
**Update:**
got the result file

https://pastebin.com/raw/Hcd18qP5

изображение

This is result thi the same input with view mode in tests

Kreeg:
I am not quite follow what kind of result is expected here...
So in this mode I can use sprite in `img src="svg#name"` elements like in symbol mode and i still can use css backgound?
The results are 
изображение
So css mode is working but `img` is showing whole svg file.
Am i wrong somewhere or it is just code isn't woring?

flobacher:
@Kreeg I usually work with symbol mode via the following markup

```
<svg viewBox="0 0 30 10">
  <use href="./path-to-my-spritesheet/svg-sprites.svg#mysymbol" />
</svg>
```
I don't think referencing a symbol from an svg works via img tag..

Kreeg:
@flobacher Ok I'll what we can do here. The problem here is in tests. There is a browser (`puppeteer` via chromium)  security problem in assessing `file://... svg` from `file://...html`.

Kreeg:
Ok. that is what I've got now 
изображение
The problem here now is with the css version. It's just differs from `svg use` version

flobacher:
@Kreeg  hm.. looks like this is because of the gradients that are also referenced via id

Kreeg:
Maybe this is related #74?

PR #99, opened 2015-07-24
Remove "." (dot). Selector already has one.

panych:
Configuration field 'prefix' should contain first selector parts including class dot. To prevent duplication we remove one from template. 
https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md#common-mode-properties


jkphl:
@panych Thanks for this one! :)

PR #116, opened 2015-10-12
Moved phantomjs from npm dependencies to devDependencies

goggenb:
Should PhantomJS really be a normal dependency and not a dev dependency?


jkphl:
PhantomJS is used for determining the dimensions of SVG files that don't have `width` / `height` / `viewBox` attributes properly set. So yes, unfortunately, PhantomJS is a regular requirement. I'd love to make this configurable, however I don't know of a way to do so. I'm open for suggestions into this direction — but removing PhantomJS as regular requirement is not an option right now. Thanks for thinking about this, however.

goggenb:
Ok, thanks for the explanation.

PR #123, opened 2015-11-15
Fix example config value

miljan-aleksic:
At least in my testing without the `%s` nothing will be prefixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ var config					= {
 		<mode> 				: {
 			dest			: "<mode>",						// Mode specific output directory
 			prefix			: "svg-%s",						// Prefix for CSS selectors
-			dimensions		: "-dims",						// Suffix for dimension CSS selectors
+			dimensions		: "%-dims",						// Suffix for dimension CSS selectors
 			sprite			: "svg/sprite.<mode>.svg"		// Sprite path and name
 			bust			: true|false,					// Cache busting (mode dependent default value)
 			render			: {								// Stylesheet rendering definitions
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ var config					= {
 		<mode> 				: {
 			dest			: "<mode>",						// Mode specific output directory
 			prefix			: "svg-%s",						// Prefix for CSS selectors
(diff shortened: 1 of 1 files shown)


jkphl:
@miljan-aleksic Thanks for reporting this. However, two things:

1.) Doesn't your patch miss the `"s"` then (i.e. `"%s-dims"` instead of `"%-dims"`)?
2.) [This line](https://github.com/miljan-aleksic/svg-sprite/blob/master/lib/svg-sprite/mode/standalone.js#L47) should make that any string for `dimensions` that doesn't contain `"%s"` is treated just like `"<whatever-your-prefix-value-is><whatever-your-dimensions-value-is>"`, and it didn't change for a long time, so I wonder why it shouldn't work (anymore).

I will have to dig into this before merging the PR. Thanks for your patience.

miljan-aleksic:
Dah... fixed. It could be my configuration went wrong somewhere, although I just followed the simple example.

jkphl:
Thanks again. As I said, I'll need to examine this further, so please bear with me.

miljan-aleksic:
No rush, keep it up and thanks :)

jkphl:
I'm sorry but I can't reproduce the behaviour you're experiencing. With this very simplistic configuration I'm getting perfectly suffixed dimension CSS classes:

```js
var config = {
    "mode": {
        "css": {
            "dimensions": "-dims",
            "render": {
                "css": true
            }
        }
    }
}
```

results in this CSS: 

```css
.svg-weather-clear {
	background: url(svg/sprite.css-5573d147.svg) 100% 0 no-repeat;
}

.svg-weather-clear-dims {
	width: 48px;
	height: 48px;
}

.svg-weather-snow {
	background: url(svg/sprite.css-5573d147.svg) 100% 5.2631578947368425% no-repeat;
}

.svg-weather-snow-dims {
	width: 48px;
	height: 48px;
}

/* ... */
```

Could you please post your complete configuration? Maybe there's another error ...

miljan-aleksic:
I can't reproduce it again neither, must be some specific configuration that reproduce it or me having a lapsus... If I run into it again will get back to you.

PR #129, opened 2015-12-18
fix typo on configuration.md

tgfjt:


jkphl:
どうもすみません, but the term *prolyfill* has been chosen [intentionally](https://au.si/what-is-a-prolyfill). ;)

jkphl:
@tgfjt One question though: Are you actively using svg stacks as described in that part of the documentation?

tgfjt:
Oh I see. sorry, my bad!

I just use symbols now.
I wonder if I should use defs or symbols or stacks... Is it stacks good??

jkphl:
@tgfjt I'd stick to symbols as well! :)

Personally I don't think stacks are the way to go. I was just wondering if anybody is using them at all. I'm planning a new major release and am thinking about dropping support completely ...

tgfjt:
I agree with dropping support.
We want simple and stable way!

PR #140, opened 2016-02-07
Few files had been mistakenly treated as a symlink being a normal file

Johnius:
While cloning the repo I have faced the problem that few `svg` files were a normal files but looked like a `symlink` so their content was treated as a pathname.

That is the reason for such an errors
> error: unable to create symlink test/fixture/svg/css/weather-clear.svg (File name too long)
error: unable to create symlink test/fixture/svg/css/weather-snow.svg (File name too long)
error: unable to create symlink test/fixture/svg/css/weather-storm.svg (File name too long)

Googling the error showed this solution
http://stackoverflow.com/questions/18411200/git-unable-to-create-symlink-file-name-too-long


Johnius:
The error is also shown in Travis CI task log here https://travis-ci.org/jkphl/svg-sprite/jobs/107670091

nikini:
I get the exact same build errors.

jkphl:
@Johnius @nikini Thanks for reporting this! However, simply removing the files / symlinks like this is not an option as I'm using them for extended tests during development. I added them to `.gitignore` so they won't be commited to GitHub again in the future. I have no idea why the symlinks suddenly became a problem — they had been there for ages ... Anyway, thanks again! :)

nikini:
@jkphl Great man :)
It's always nice to see an active repository being updated regularly!

jkphl:
@nikini Thanks — you're welcome! :)

PR #141, opened 2016-02-09
Problem when dimensions are not rounded properly

nikini:
I noticed the problem by adding precision `0` to the `shape.dimension`. The values were not rounded in the stylesheet. Then I looked around and it appeared that when the values are coming from the bounding box or from phantomjs they are not rounded at all. 

When loading the style into chrome I had issues with some svgs which were slightly taller. I get that the svgs should never have a width/height with decimals precision, but the plugin should also handle that.

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -475,8 +475,8 @@ SVGShape.prototype._determineDimensions = function(cb) {
 	
 	// Try to use a viewBox attribute for image determination
 	if (this.viewBox !== false) {
-		this.width				= this.viewBox[2];
-		this.height				= this.viewBox[3];
+		this.width				= this._round(this.viewBox[2]);
+		this.height				= this._round(this.viewBox[3]);
 	}
 
 	// If the viewBox attribute didn't suffice: Render the SVG image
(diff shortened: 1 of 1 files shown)


nikini:
What's up with that build by the way? How can I fix the build? I only changed that file, nothing else

jkphl:
@nikini Thanks for the PR. It's most likely not your fault that the build is failing at the moment. There have been some issues with GitHub & Travis during the last couple of days (as far as I have heard), and also I'm about updating some problematic dependencies. Don't worry. I'll take care of your PR as soon as I finished the updates.

nikini:
When can you merge?

jkphl:
@nikini I'm sorry but
1. I'm on holidays right now and don't have access to my dev environment,
2. tests are failing at the moment because some dependencies need to be updated which makes some major rework necessary and
3. I totally lack the time for working on side projects like this at the moment.
I'm sorry but I can't predict right now when I'll find the time to get this out of the door.

nikini:
Any news on this?

jkphl:
No news, sorry. I won't be able to work on side projects like this one until end of April. Please feel free to provide PRs that fully integrate the new dependency versions.

PR #148, opened 2016-03-30
Change phantomjs into phantomjs-prebuild

davesnx:
Hi @jkphl 

I just change the dependency of `phantomjs` into `phantomjs-prebuild`
Because I have random errors on Bitbucket 403 and 429 (too many requests), actually because `phantomjs` is downloading itself or a dependency that they need from a repo on bitbucket and bitbucket are responding with some errors.
and this stuff is breaking my build :cry: 

Source: [phantomjs#deprecated](https://www.npmjs.com/package/phantomjs#deprecated)

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
 		"svgo": "0.6.1",
 		"cssom": "^0.3.0",
 		"css-selector-parser": "^1.1.0",
-		"phantomjs": "^1.9.19",
+		"phantomjs-prebuilt": "^2.1.7",
 		"cssmin": "^0.4.3",
 		"mustache": "^2.2.1",
 		"js-yaml": "^3.5.2",


davesnx:
In case that you don't like this,

We can use a CDN for download the `phantomjs` and maybe can solve this problem.

So the solution is add the env in `.npmrc`  `phantomjs_cdnurl`=`http://cnpmjs.org/downloads`
**info**: [Deciding Where To Get PhantomJS](https://www.npmjs.com/package/phantomjs)

jkphl:
Thanks for the suggestion @davesnx. I will review it as soon as possible, but I won't have time to work on side projects like *svg-sprite* until end of April. The Travis build fails because of several dependencies that need to be updated and tested and I simply don't find the time to work on it due to general overload. Please bear with me.

jkphl:
Fixed with commit https://github.com/jkphl/svg-sprite/commit/50af8d5a9f45055a5b969a8c496827008c886272

PR #159, opened 2016-05-18
Fix incorrect lookup for shape-transform-* config

middric:
Documentation states that you can provide transform config json files using the argument "--shape-transform-*". The code actually looks for "--transform-*". Change fixes this to match the documented behaviour.

### DIFF
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -150,9 +150,9 @@ var transform						= ('' + config.shape.transform).trim();
 config.shape.transform					= [];
 (transform.length ? transform.split(',').map(function(trans){ return ('' + trans).trim(); }) : []).forEach(function(transform){
 	if (transform.length) {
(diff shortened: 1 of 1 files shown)


jkphl:
Thansk for reporting @middric! I'm not entirely sure if this fix is correct — there used to be one top-level `transform` property (this is probably what the code still refers to), but in the meantime there are two of them (`shape-transform` and `svg-transform`). Did you try passing in JSON file for configuring e.g. SVGO? Did it work? Then it's probably the correct transform property you picked ... Thanks for feedback.

middric:
Yep I tried it for shape-transform and svgo. Worked correctly! The block assigns the values from the json file to "config.shape.transform"

jkphl:
Awesome, thanks for confirming. I merged the PR, will be part of the next release! Thanks again.

PR #164, opened 2016-05-31
fixed jkphl/svg-sprite#163

daanvanham:
actually check the config if the user wants the dimension attributes set before setting the attributes

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -405,10 +405,12 @@ SVGShape.prototype.getDimensions = function() {
  * @return {SVGShape}			Self reference
  */
 SVGShape.prototype.setDimensions = function(width, height) {
-	this.width				= this._round(Math.max(0, parseFloat(width, 10)));
-	this.dom.documentElement.setAttribute('width', this.width);
-	this.height				= this._round(Math.max(0, parseFloat(height, 10)));
-	this.dom.documentElement.setAttribute('height', this.height);
+	if (this.config.dimension.attributes) {
+		this.width				= this._round(Math.max(0, parseFloat(width, 10)));
+		this.dom.documentElement.setAttribute('width', this.width);
+		this.height				= this._round(Math.max(0, parseFloat(height, 10)));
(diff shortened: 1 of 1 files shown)


jkphl:
@daanvanham Thanks for your "fix". However, you seem to assume that there is an option `config.<mode>.symbol.dimension.attributes` — which there isn't. Is it possible that you are looking for `shape.dimension.attributes` instead? Did you try to use that one?

daanvanham:
My current config, not working without the "fix", but working after the changes.

```
{
	mode: {
		symbol: {
			dest: '',
			sprite: './sprite.svg'
		}
	},
	shape: {
		id: {
			generator: '%s-icon'
		},
		dimension: {
			attributes: false
		}
	}
}
```

It did not result in the expected behaviour and thus I made the changes in the module, which let to not having `width` and `height` attributes on the `<symbols>`.

daanvanham:
Any news on this PR?

jkphl:
@daanvanham Firstly, I tried your exact configuration, but I'm not able to reproduce the behaviour you describe. As far as I can tell, both the `width` and `height` attribute shouldn't be available on the `<symbol>` elements as they are [filtered out](https://github.com/jkphl/svg-sprite/blob/master/lib/svg-sprite/mode/symbol.js#L60) by the [list of allowed attributes](https://github.com/jkphl/svg-sprite/blob/master/lib/svg-sprite/mode/symbol.js#L17) for the symbol sprite. As the [documentation states](https://github.com/jkphl/svg-sprite/blob/master/docs/configuration.md#shape-dimensions), the `shape.id.dimensions` option is only considered for the [defs](https://github.com/jkphl/svg-sprite/blob/master/lib/svg-sprite/mode/defs.js#L56) and [stack](https://github.com/jkphl/svg-sprite/blob/master/lib/svg-sprite/mode/stack.js#L72) sprites. Your patch shouldn't be necessary. Please double-check your configuration once more.

I will close this PR for now as I can't solve a problem that is not reproducible. As soon as I can, I'll happily reconsider it.

PR #165, opened 2016-06-10
Added config option to CLI that uses JSON file for spriter configuration

Laruxo:


jkphl:
@Laruxo Thanks again for this great contribution!

I needed to adapt your PR a little, as in certain situations there was a conflict between the default values and the values in the external config file. It would be ideal to use the config file as a base and then consume the additional command line options as an overlay, but unfortunately this doesn't really work, as there are a lot of default values that **must** be set. When merging the default values onto the config file, I can't always decide which one is a "must have" setting or one that should be overridden by the config file. So I basically turned it the other way round so that the config file is merged onto the default values. This isn't perfect either, but at least all necessary are set this way ...

Anyway, thanks again! :)

Laruxo:
@jkphl No problem :)

That was what I wanted to do at first, however as you, I also reached a conclusion, that with current implementation of default values that is not possible. I don't really like how it needs to check both CLI arguments and values from config file, but I see no other way to implement this without doing major rewrite.

Maybe in the future, if you do a refactoring of that part, default values could be defined as an object into which config object would be merged. Then, if there are any CLI arguments they would simply overwrite properties of that merged object.

Anyway, keep up the good work with this module! :)

jkphl:
@Laruxo Thank you! :)

Yeah, this would be the only feasible way to go I guess. Right now, however, I'm using [yargs](http://yargs.js.org/) which does all the argument stuff — without any chance to interfere ... (Well, in fact I tried disabling the default values when a `--config` argument is present, but some of the default values are mandatory, obviously ...)

But still, at least we got it a little bit better! Thanks for that. :)

Issue #133, opened 2016-01-03
Invalid call mixin in Less 


mrmlnc:
Hi, 

I wrote the following configuration:

```js
mode: {
  sprite: 'sprite.svg',
  css: {
    dest: 'styles/'
    common: 'icon',
    prefix: '.icon-',
    dimensions: '-wh',
    sprite: '../images/sprite.svg',
    bust: false,
    mixin: 'sprite-icon',
    render: {
      less: true
    }
  }
}
```

Gulp plugin generated for me this code:

```less
.sprite-icon() {
	background: url("../images/sprite.svg") no-repeat;
}
.icon {
	sprite-icon();  // here is missing the dot
}
// More icons
// ...
```

If I specify a dot in the property `mode.<mode>.mixin` (`mixin: '.sprite-icon'`), then:

```less
..sprite-icon() { // here two dots
	background: url("../images/sprite.svg") no-repeat;
}
.icon {
	.sprite-icon();
}
// More icons
// ...
```

mrmlnc:
Good sounds!

jkphl:
@mrmlnc Yep, should be fixed as of 1.2.18 ;)

PR #187, opened 2016-08-18
Update sprite.less to avoid compilation error on mixin call

kazmiekr:
Adds a dot before calling the less mixin otherwise this will generate syntax errors.

I'm using the following configuration in grunt-svg-sprite:

```
target: {
		src: '../svgs/*',
		dest: '.',
		options: {
			shape: {
				id: {
					generator: generatorFunc
				},
				dimension: {
					maxWidth: 76,
					maxHeight: 76
				},
				spacing: {
					padding: 1
				}
			},
			mode: {
				css: {
					bust: false,
					mixin: 'svg-mysvgs-common',
					sprite: '../img/Icons',
					prefix: '.myPrefix-',
					dimensions: true,
					render: {
						less: {
							dest: '../src/less/somelessfile.less'
						}
					}
				}
			}
		}
	},
```

will generate something like:
```
.svg-mysvgs-common() {
	background: url("../img/whatever.svg") no-repeat;
}

.myPrefix-some-svg {
	svg-trait-small-common();
	background-position: 16.666666666666668% 0;
	width: 50px;
	height: 50px;
}
```

in the output `svg-trait-small-common()` will cause a less compiler error. It needs the '.' before it.

This looks similar to a previous issue #133 

Thanks so much for the hard work in this library!

### DIFF
--- a/tmpl/css/sprite.less
+++ b/tmpl/css/sprite.less
@@ -12,7 +12,7 @@
 
 {{/hasMixin}}{{#shapes}}{{#selector.shape}}{{expression}}{{^last}},
 {{/last}}{{/selector.shape}} {
-	{{^hasCommon}}{{#hasMixin}}{{mixinName}}();{{/hasMixin}}{{^hasMixin}}.{{commonName}}();{{/hasMixin}}
+	{{^hasCommon}}.{{#hasMixin}}{{mixinName}}();{{/hasMixin}}{{^hasMixin}}.{{commonName}}();{{/hasMixin}}
 	{{/hasCommon}}background-position: {{position.relative.xy}};{{#dimensions.inline}}
 	width: {{width.outer}}px;
 	height: {{height.outer}}px;{{/dimensions.inline}}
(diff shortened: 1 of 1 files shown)


jkphl:
@kazmiekr Thanks for reporting this. However, with your patch applied the tests are now failing (see [Travis results](https://travis-ci.org/jkphl/svg-sprite/builds/153359835)). I find the following LESS file created:

```css
.svg-common() {
        background: url("svg/css.vertical-b87f0584.svg") no-repeat;
}

.svg-weather-clear {
        ..svg-common();
        background-position: 0 0;
        width: 48px;
        height: 48px;
}

/* ... */
```

As you see there are now two dots in front of the mixin call, whereas this has only been one prior to your patch. Could you please double-check that you are not using a pre-1.2.18 version of *svg-sprite*?

kazmiekr:
Updated the PR with a new commit, build seems to be passing now, I copied the dot into the wrong spot from where I had it in my local revision. Let me know what you think. Thanks!

jkphl:
Works now, thanks! Merged. ;)

PR #195, opened 2016-10-11
Fix regexp to handle quotes in url()

sshev:
I had an error that in some cases original `clip-path` attribute is absent in resulting sprite, in particular it's happening for attributes like `clip-path="url('#someID')"`

after debugging turned out that `svg-sprite` incorrectly handles quotes in `url()` - see this [jsbin](http://jsbin.com/bulinidota/edit?js,console)

so here's fix for this issue


jkphl:
Looks very reasonable, thank you! :)

danbovey:
Can you publish this into a new version on npm please?

jkphl:
@danbovey Currently the tests are failing and I'll have to update quite some modules before I can push out a new release. However, I plan to do so around end of March if nothing unexpected is going to happen.

PR #204, opened 2016-11-18
fix broken sprite urls in example html files

c0defre4k:


jkphl:
@c0defre4k I know it's been a long time — and I'm sorry for that! — but I just wanted to review this PR. It seems to me, however, that you must be using something in a way that it's not intended for?! The original code with `{{{example}}}` just works fine in all my tests, I tried all of the relevant sprite formats. Is this still an issue for you? If not, I'd rather dismiss this PR. I'm about moving on and doing a complete rewrite, so this will probably not be relevant in the future anyway ... Thanks for your feedback!

jkphl:
It turns out this is the last issue I'd need to finish to reach the final milestone for the 1.x branch and I cannot reproduce your problem, everything seems to just work fine. I will therefore close this issue for now. Please get back to me if you still encounter these problems though. Thanks!

PR #206, opened 2016-12-14
Make docs for the config object consistent

brendanfalkowski:
Added `svg.<node>` before each property in the table so it's consistent with the docs for `mode.<node>` and `shape.<node>`.


jkphl:
Thank you! :)

PR #219, opened 2017-02-15
Solve issue with showing incorrect month in log

Pudge601:
The `getMonth` method returned a zero-based integer value for the month (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth), which would therefore be showing 1 month behind what the actual month is.

Example output from today (February 15th) which shows the issue:
screen shot 2017-02-15 at 15 04 51

The `Date.format` provides the same formatted date which the timestamp method was previously building manually.


jkphl:
@Pudge601 Good catch, thanks! :+1:

PR #258, opened 2018-02-10
Use more recent version of svgo

vincentbernat:
The optimize() function now returns a promise. Adapt the code to
behave correctly in this case.

We could also just keep the older svgo dependency.


jkphl:
Thanks! Had to update the test fixture as well ...

PR #261, opened 2018-03-17
Enforce linting

TheDancingCode:
Fixes JSHint errors that went uncaught. Adds linting as a `pretest` task, to avoid further errors.

I recommend switching from JSHint to ESLint and its `eslint:recommended` rules.


jkphl:
@TheDancingCode Thanks!

I agree that ESLint would be the better choice nowadays (and I'm using ESLint in other projects). However, I'm planning a complete rewrite of *svg-sprite* and I probably wouldn't switch prior to that (I'm utterly lacking the time for this all, though ...)

PR #284, opened 2019-04-23
Replace phantomjs with puppeteer

### LINKED ISSUE #254: Replace PhantomJS with Puppeteer
[I see](https://github.com/jkphl/gulp-svg-sprite/issues/69) [that](https://github.com/jkphl/svg-sprite/issues/100) [many](https://github.com/jkphl/svg-sprite/issues/239) [people](https://github.com/jkphl/svg-sprite/issues/211) [have](https://github.com/jkphl/svg-sprite/issues/222) [problems](https://github.com/jkphl/svg-sprite/pull/244) with the complexity of installing PhantomJS.

Also, the project is officially dead. Do you think it's a good idea to replace PhantomJS with [Puppeteer](https://github.com/GoogleChrome/puppeteer)?
Find out the real dimensions of SVG without width and height there is not difficult.
Puppeteer is entirely node.js-compatible, might correctly installed almost everywhere. It indeed is more compatible than PhantomJS. The browser in it can be run with parameters for greater compatibility (disabling hardware acceleration, for example).

Do you think it's a good idea to replace PhantomJS with [Puppeteer](https://github.com/GoogleChrome/puppeteer)?

mojoaxel:
Hi @jkphl :wink: 

This is an attempt to get rid of the phantom.js dependency (see #254):
* replaced `phantomjs` with `puppeteer`
* replaced `svg2png` with `sharp`.

So far I have encountered the following problems:

1. All PNG renderings with puppeteer differ very slightly from the originals from phantom. Because the images are not pixel-perfect I had to replace all expected images for the tests to be green again. Because a pixel-perfect for PNG renderings can not be ensured **I would consider this a breaking change**.
2. The `sharp` library prints a "CSS parsing error" while saving a SVG as PNG. I'm not sure were exactly this problems occurs and why. I'm going to create an issue with the sharp library to see if this can be solved.
3. The test case `view\png\css.horizontal.mixed.png` is still failing.

I would consider this a first proof of concept and would be happy If you folks could test this branch an see if it still works in you setup.

Fixes #254

TODO:

- [x] split the looks-same patch. I gave this a go in but I'm not 100% sure it's OK yet - Opened #417
- [ ] see if we really need sharp; we should be able to write the PNG image from puppeteer and just compare it?
- [ ] update puppeteer to the latest version
- [ ] make puppeteer optional
- [ ] move the newly added puppeteer code to a new file and reduce duplication in tests
- [ ] fix the intermittent test failures


jkphl:
@mojoaxel Hi Alex! Awesome, thank you! I will review this as soon as possible, but it will take me some time due to general overload. Please bear with me. :)

lib/svg-sprite/shape.js
```diff
@@ -20,9 +20,7 @@ xpath						= require('xpath'),
 cssom						= require('cssom'),
 CssSelectorParser			= require('css-selector-parser').CssSelectorParser,
 csssel						= new CssSelectorParser(),
-execFile					= require('child_process').execFile,
-phantomjs					= require('phantomjs-prebuilt').path,
-dimensionsPhantomScript		= path.resolve(__dirname, 'shape/dimensions.phantom.js'),
+puppeteer = require('puppeteer'),
```
ogonkov: Changes shouldn't interrupt code style of the file

sp90:
@jkphl are we doing this?

mojoaxel:
> @jkphl are we doing this?

let me know it I can help. I met @jkphl last week and this PR came back to mind. I'll try to have a look this week and see if I can improve it. I still think it's worth a breaking change.

jkphl:
Sorry all. I'm totally swamped with tasks for months now and I just don't find any time to take care for my OS side projects.

@mojoaxel very thankful for your efforts! Maybe we should have a personal chat before you go a head and do something about this? You can catch me in the office whenever you find some time to call me.

test/svg-sprite.js
```diff
@@ -390,13 +400,10 @@ describe('svg-sprite', function () {
 
                     capturePuppeteer(preview, previewImage, function (error) {
                         should(error).not.ok;
-                        imageDiff({
-                            actualImage: previewImage,
-                            expectedImage: path.join(__dirname, 'expected', 'png', 'css.html.png'),
-                            diffImage: path.join(__dirname, '..', 'tmp', 'css', 'png', 'css.html.diff.png')
-                        }, function (error, imagesAreSame) {
+                        looksSame(previewImage,path.join(__dirname, 'expected', 'png', 'css.html.png'),
+                        function (error, r) {
```
ogonkov: what is `r`?

mojoaxel: Maybe @bung87 can answer this?

mojoaxel: I think "r" is short for "result". I'll improve..

mojoaxel:
Thanks to @bung87 we could improve this further. Now there are still 3 tests failing and we should talk about on how to handle this!

## creates visually correct sprite with - packed layout

#### expected
image

#### current
image

## creates visually correct sprite (horizontal.mixed)

#### expected 
image
#### current
image

mojoaxel:
> creates visually correct sprite with - packed layout

Strangely I get the same failed tests also with `master`. So it looks like this problem has nothing to do with this PR.

test/svg-sprite.js
```diff
@@ -14,19 +14,17 @@
  */
 
 var fs = require('pn/fs'); // https://www.npmjs.com/package/pn
-var svg2png = require('svg2png');
+var sharp = require('sharp');
```
ogonkov: Why did you use `sharp`?

package.json
```diff
@@ -48,8 +48,8 @@
     "mkdirp": "^0.5.1",
     "mocha": "^5.2.0",
     "mustache": "^3.0.0",
-    "phantomjs-prebuilt": "^2.1.16",
     "prettysize": "^1.1.0",
+    "puppeteer": "^2.0.0",
```
ogonkov: If phantom/puppeter is used only to find SVG dimensions when it missing, it could be moved to `optionalDependencies`

package.json
```diff
@@ -60,18 +60,18 @@
...
     "coveralls": "^3.0.2",
-    "image-diff": "^2.0.0",
     "istanbul": "*",
     "jshint": "^2.9.6",
     "less": "^3.8.1",
+    "looks-same": "^7.2.2",
     "mocha": "",
     "mocha-lcov-reporter": "*",
-    "node-sass": "^4.9.3",
+    "node-sass": "^4.13.0",
```
ogonkov: `node-sass` is replaced by `sass` (this could affect css output and tests results)

mojoaxel: ? It looks to me like `node-sass` is still under active development.

ogonkov: node-sass is C lib, sass is officially support Dart implementation that is distributed as `sass` package

mojoaxel:
I rebased on master and updated the dependencies.
There are still the two open issues with the test that should be discussed.

package.json
mojoaxel: Feel free to provide a seperate pull-request that replaces `node-sass` with `sass`! 
I think this is not in the scope of this PR.

jkphl:
@mojoaxel Regarding the failing tests:

* The Zebra problem could be what we talked about earlier (screenshot must not be taken immediately as browser might need a bit to process CSS within the SVG file)
* The problem with the order of the packed sprite somehow seems to be related to the node version. See the [latest Travis runs](https://travis-ci.org/jkphl/svg-sprite/builds/638875146): Node including v10 is fine, v11+ fails.

I'll look into this as soon as possible, maybe next weekend.

mojoaxel:
>  The Zebra problem could be what we talked about earlier (screenshot must not be taken immediately as browser might need a bit to process CSS within the SVG file)

Sadly this is not a timing issue. [This test](https://github.com/jkphl/svg-sprite/blob/03073deb4c8b81256afc2ad90eea07ee229bd47f/test/svg-sprite.js#L920-L930) just renders a yaml layout as svg, than saves it as png and compares it to an expected result. There is no puppeteer here.

This looks to me like this problem is caused by `sharp`. The SVG file [zebra.svg](/jkphl/svg-sprite/blob/master/test/fixture/svg/css/zebra.svg) is quite "challenging": It has embedded CSS styles that include multiple @media-queries for different screen sizes. But what if the svg-image is just rendered into a png using `sharp`?? What is the screen-size there?

Maybe this is not a problem! It should be ok to just replace the zebra in this test-case with another simpler SVG.

What do you guys think?

jkphl:
@mojoaxel I don't think it's appropriate to alter / simplify the test just to make it align with the (new) reality. Again: The tests used to work fine (and still do) with up until Node 11. There's a problem, yes, but the solution has to fix the code, not the fixture ...

mojoaxel:
> I don't think it's appropriate to alter / simplify the test just to make it align with the (new) reality.

Obviously you are right in general, but in this case it is more difficult. 
What is this test for? Why, exactly is the old implementation returning a greenish zebra? If you ask me this is just random and has no meaning at all for this test case (rendering a given a layout). With test cases like this, that test multible things at the same time and the outcome is kind of random it is impossible to make meaningful changes.

Maybe (and I doubt that) it is useful to discuss and test how SVG should handle media-queries in embedded CSS but this should IMHO be done in a seperate test case and not within multible other tests.

The question stands: What is the expected outcome if you render an SVG mit embedded css-media-queries as PNG? What if you export the same SVG using a vector-tool like Inkscape? Media queries only make sense in a browser-context not within a rendering engine!

> The tests used to work fine (and still do) with up until Node 11.

This is only the case with the "packed layout" tests. The "tebra-test" always break.
Why the "packed layout" behaved different on newer node versions is really interesting! I have no idea so far.

Antonio-Laguna:
I have been having some issues with phantom and we've realised it was also slowing down our builds. I took your PR for a spin and noticed a difference in our tests.

Input SVG was 
```
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
<rect x="10" y="10" height="100" width="100" style="stroke:#ff0000; fill: #0000ff"/>
</svg>
```

Got like 

```
<symbol viewBox="0 0 300 150" id="one" xmlns="http://www.w3.org/2000/svg">
<rect x="10" y="10" height="100" width="100" style="stroke:#ff0000; fill: #0000ff"/>
</symbol>
```

While the previous version was

```
<symbol viewBox="0 0 100 100" id="one" xmlns="http://www.w3.org/2000/svg">
<rect x="10" y="10" height="100" width="100" style="stroke:#ff0000; fill: #0000ff"/>
</symbol>
```

`viewBox` is different which makes the rendered version smaller. Thought it could be the `rect` since I have a few other SVGs for the test and the others aren't affected. Replaced the `rect` with a `path`and the issue was the same. Not sure what the issue is though.

mojoaxel:
> I took your PR for a spin and noticed a difference in our tests.

@Antonio-Laguna Thanks for the feedback! I need to look into that.

.gitignore
```diff
@@ -7,3 +7,4 @@
 /tmp
 /FEATURES.md
 /dev-scripts/
+.vscode/launch.json
```
MartijnCuppens: If we're going to add this, shouldn't we just exclude the whole `.vscode` folder?

lib/svg-sprite/shape.js
```diff
@@ -20,9 +20,7 @@ xpath						= require('xpath'),
 cssom						= require('cssom'),
 CssSelectorParser			= require('css-selector-parser').CssSelectorParser,
 csssel						= new CssSelectorParser(),
-execFile					= require('child_process').execFile,
-phantomjs					= require('phantomjs-prebuilt').path,
-dimensionsPhantomScript		= path.resolve(__dirname, 'shape/dimensions.phantom.js'),
+puppeteer				= require('puppeteer'),
```
MartijnCuppens: This needs an additional indent
```suggestion
puppeteer					= require('puppeteer'),
```

.gitignore
mojoaxel: fixed

lib/svg-sprite/shape.js
mojoaxel: fixed

mojoaxel:
While the debate here goes on I published the phantomjs and svg2png free version on npm:
[@fullstax/svg-sprite](https://www.npmjs.com/package/@fullstax/svg-sprite)

gabbersepp:
> While the debate here goes on I published the phantomjs and svg2png free version on npm:
> [@fullstax/svg-sprite](https://www.npmjs.com/package/@fullstax/svg-sprite)

Hey I just tested the lib and I encountered differences between our old sprite and the new one:

image

But there is no need to provides fixes. We just continue to use the old lib.

MartijnCuppens:
@gabbersepp, that's an optimalisation, those ones and zeros are booleans, that's why the spaces can be removed.

gabbersepp:
> @gabbersepp, that's an optimalisation, those ones and zeros are booleans, that's why the spaces can be removed.

Good to know. Is there a possibility to disable this? I just want to see if both sprites are absolutly equal.

MartijnCuppens:
> Good to know. Is there a possibility to disable this? I just want to see if both sprites are absolutly equal.

No clue, but maybe you can optimise the old sprite file with svgo?

mojoaxel:
@gabbersepp My changes introduce a major change! It is naturally not a perfect drop-in replacement. The results should be optically and functionality but not binary identical!

gabbersepp:
> @gabbersepp My changes introduce a major change! It is naturally not a perfect drop-in replacement. The results should be optically and functionality but not binary identical!

Thanks for your work! I did an "optical" comparization and everything seems good.

package.json
Ptico: Fully agree, viewbox is a good practice and forcing 200MB downloads for everyone who follow good practices is bad. What about to throw a meaningful message when puppeteer is needed but not installed?

Jordan-Hall:
@mojoaxel i dropped your plugin replacement and puppeteer came back with an error file://dummy.svg not found

XhmikosR:
@mojoaxel I'd be interested in this after we land some of the other open PRs first to drop Node.js 8.x and sort out some things.

The test failures are something I haven't figured them out. Not sure if it's the test code, phantomjs or something in Node.js but happens in all Node.js versions >= 10.

Also, it'd would be indeed nice to lazyload puppeteer, make it an optional dependency and throw a nice error if someone tries to use the feature that needs puppeteer.

package.json
```diff
@@ -47,9 +47,10 @@
     "lodash.pluck": "^3.1.2",
     "mkdirp": "^0.5.1",
     "mustache": "^4.0.0",
-    "phantomjs-prebuilt": "^2.1.16",
     "prettysize": "^2.0.0",
     "svgo": "^1.3.2",
+    "puppeteer": "^2.0.0",
+    "should": "^13.2.3",
```
XhmikosR, line 53: This needs to be moved to devDependencies

XhmikosR:
In fact, we could land the looks-same patch much sooner if you could make a PR for that separately.

test/svg-sprite.js
```diff
@@ -131,21 +160,29 @@ function compareSvg2Png(svg, png, expected, diff, done, msg) {
...
-                        diffImage: diff
-                    }, function (err, imagesAreSame) {
-                        should(err).not.ok;
-                        should.ok(imagesAreSame, msg);
-                        done();
-                    });
-                })
-                .catch(ecb);
+        .then(function(buffer) {
+            sharp(buffer).toFile(png, function(err) {
```
XhmikosR, line 164: I was wondering, can't we use `readFile` without an `enconding` set (thus returning a Buffer) and `writeFile`? Do we need sharp?

XhmikosR:
Rebased this. We need to finish with #398 first.

What would help is if:

1. someone could split the looks-same patch. I gave this a go in https://github.com/jkphl/svg-sprite/commit/8eea02e332a739110e3ddd4e7d79a5e1ff2340b2 but I'm not 100% sure it's OK yet
2. make puppeteer optional
3. see if we really need sharp
4. this PR still doesn't seem to fix the Node.js > 10 test failures
5. update puppeteer to the latest version

XhmikosR:
Rebased and updated the TODO in the PR description. Any help would be appreciated to sort the TODO so that we can land this.

lib/svg-sprite/shape.js
```diff
@@ -513,20 +511,60 @@ SVGShape.prototype._determineDimensions = function(cb) {
...
+			.then(function() {
+				return _browser.close();
+			})
+			.then(function() {
+				cb();
+			})
+			.catch(function(err) {
+				console.error(err);
+			});
+		} catch(err) {
```
ert78gb, line 564: There is no synchronous code in the try-catch block, so it will catch nothing. Have to move the `cb(new Error(err));` to the line 562

lib/svg-sprite/shape.js
```diff
@@ -513,20 +511,60 @@ SVGShape.prototype._determineDimensions = function(cb) {
...
+				return _page.evaluate(function(node) {
+					var rect = node.getBoundingClientRect();
+					return { width: rect.width, height: rect.height };
+				}, nodeHandle);
+			})
+			.then(function(bounds) {
+				that.width = that._round(bounds.width);
+				that.height = that._round(bounds.height);
+				return _nodeHandle.dispose();
+			})
```
ert78gb, line 554: I think can be simplify the 3 function into 2 and eliminate the `_nodeHandle` variable.
```suggestion
			.then(function() {
				return _page.evaluate(function() {
					const rect = $('svg').getBoundingClientRect()
					return { width: rect.width, height: rect.height };
				});
			})
			.then(function(bounds) {
				that.width = that._round(bounds.width);
				that.height = that._round(bounds.height);
			})
```

lib/svg-sprite/shape.js
```diff
@@ -513,20 +511,60 @@ SVGShape.prototype._determineDimensions = function(cb) {
...
-	        	cb(null);
-	        } else if (stderr.length > 0) {
-	            cb(new Error(stderr.toString().trim()));
-	        } else {
-	            cb(new Error('PhantomJS didn\'t return dimensions for "' + that.name + '"'));
-	        }
-	    });
+		var _browser, _page, _nodeHandle;
+		try {
+			puppeteer.launch()
```
ert78gb, line 516: Start a new browser instance is a very expensive operation. I think worth to implement a singleton for the browser instantiation  and dispose the the browser object when the `svg-sprite` finish the job.

lib/svg-sprite/shape.js
```diff
@@ -513,20 +511,60 @@ SVGShape.prototype._determineDimensions = function(cb) {
...
+					return { width: rect.width, height: rect.height };
+				}, nodeHandle);
+			})
+			.then(function(bounds) {
+				that.width = that._round(bounds.width);
+				that.height = that._round(bounds.height);
+				return _nodeHandle.dispose();
+			})
+			.then(function() {
+				return _browser.close();
```
ert78gb, line 556: If there is any error in the process the chromium will not exit and will consume the resources.
If you implement the singleton pattern then it can be a solution or close the browser in the catch block too

XhmikosR:
@ert78gb might be easier if you made a new PR with your fixes. See also the TODO in the PR's first post.

ert78gb:
Yesterday I wanted to investigate the #421 but on Mac, not every test pass. If I check the diff png I don't see any difference. I ran the #417 on my machine and every test passed.
Don't you think it worth adding windows and mac to the test matrix? 

If you merge #417 I will create a new PR.

XhmikosR:
@ert78gb that's normal because phantomjs has known issues with DPI and produces different images, at least on Windows. My plan is to at least add Windows on CI after we remove puppeteer.

Temporarily you could use the CI, that's what I've been using all this time too. When tests fail, the images are uploaded on CI.

So, the puppeteer change doesn't need to be tied to looks-same, they are 2 different things AFAICT.

#417 isn't ready either AFAICT; it has its own TODO.

ert78gb:
Wait for CI is increase the development time, but I will try. CI is great to run the test matrix, but when the tests fail in local ... :(

If you have TODO for a PR and it is not a blocker I think worth creating a separate issue. Do everything in 1 PR is not efficient.

I think this refactoring period worth doing many small PR, because the review is much easier.
I am a little bit worried about this project because wait 2-3 week for a PR merge not good.

If I see well not @XhmikosR the bottleneck.

XhmikosR:
Well, we are all volunteers, so you know. I explained to you the current TODOs and I have already broken this PR into 2 parts.

Unfortunately, I can't help you with CI, it is what it is. And it was much worse before my recent patches; they never passed on Node.js > 10. Although, I'm tempted to just drop Node.js 10 support too being that it goes EOL at the end of April. I'd still like to have a stable sorting algo, but it's one less thing to worry about.

PR #320, opened 2020-01-30
test: only save .diff if change exists

mojoaxel:
Only save `*.diff.png` files if images are not the same.
This helps wile writing tests to see if and where something changed.

I also renemed two diff files path that looked like copy-paste errors to me.


XhmikosR:
@mojoaxel I think it'd be better if you could also switch to https://github.com/gemini-testing/looks-same or https://github.com/mapbox/pixelmatch since image-diff is deprecated.

BTW we also maintain https://github.com/twbs/svg-sprite and I have made #378 (I know it's become huge, but I haven't received any reply from the maintainer so I keep adding stuff to that branch)

XhmikosR:
@mojoaxel can you split the test renaming to a new PR please? I will start landing a few patches from my branch and it indeed looks like a typo.

mojoaxel:
@XhmikosR Thanks for looking into all this!! I'll try to look into the issues in the next days.

XhmikosR:
So, I was thinking after switching to look-same we should handle this. But ideally we shouldn't write the diff image if there's no need to.

XhmikosR:
Closing this since image-diff is replaced with looks-same on master. Feel free to make a new PR against master @mojoaxel :)

PR #321, opened 2020-01-30
test: replace the zebra with a simple gene image

mojoaxel:
The zebra image has embedded css with size-dependent media-queries and is therefore not ideal to test other functionality like layouts. The idea of this PR is to remove complexity in the existing tests so that complex changes (like #284) can be done more easily.

I therefore replaced the zebra image with a more simple image (Image-License is CC0 1.0: https://svgsilh.com/image/718905.html) and updated the expected test-images.

Tests with styled SVG (like the zebra) will be part of another pull-request.


XhmikosR:
For what is worth, I reported the issue on svgo: https://github.com/svg/svgo/issues/1359

jkphl:
@XhmikosR This is a tough one, and I had [discussions with @mojoaxel](https://github.com/svg-sprite/svg-sprite/pull/284#issuecomment-576449874) about it earlier. I'd be interested in your opinion. To give you a short intro:

As you will know, SVGs may contain CSS including media queries. The media queries may respond to the current screen size, making the the SVG a "responsive SVG". Under certain conditions (= screen size) a certain set of CSS rules should apply, others shouldn't apply. The Zebra image is such an SVG, basically only switching colors between screen sizes. And as I said, under certain conditions (which are clearly controlled by the test case) a certain visual output should be generated. Of course it's not *svg-sprite*'s job to implement the SVG responsiveness — this has to be done by the browser rendering the SVG. **BUT** *svg-sprite* has to make sure it doesn't destroy that functionality while adding the shape to a sprite, hence I think it's relevant for testing.

Alex introduced Puppeteer which kinda had a problem with the Zebra (iirc), rendering false results. I understand it's a difficult situation, but I'm not sure whether simply removing / replacing the Zebra can be considered a "solution". I still think it's important to ensure that *svg-sprite* is not the cause problems. What do you think?

XhmikosR:
@jkphl there's also #322. While I agree that we should test this case too, it shouldn't block us.

That being said, in #284 tests mostly pass without removing the file. The are some failures but I'm not sure if it's due to some async issues or the outdated puppeteer version. We should discuss these there.

Let's put this on hold for now until we make the move to puppeteer.

XhmikosR:
@jkphl actually this would sort of fix the result test images in #403. The upstream issue is still not fixed.

XhmikosR (approved):

PR #325, opened 2020-02-11
Fix missing shape names when adding input files with a relative path

Sidnioulz:
This commit ensures that input files' path is resolved *before* it's split into a base path and full path. Indeed, Shape's constructor builds shape names as follows:

```
this.name = this.source.path.substr(this.source.base.length + path.sep.length);
```

With the current code, only the source file's base is resolved, and so, `source.base` is longer than `source.path`, and the whole path gets destroyed, resulting in `this.name` being an empty string.

With the proposed change, shape names are now correct even when input files have a relative path.

Note that the `else` branch was used to ensure consistent behaviour with previous code; though I'm unsure what the purpose of `file.base = path.resolve(file.base);` is.

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -127,9 +127,9 @@ SVGSpriter.prototype.add = function(file, name, svg) {
             path: file,
             contents: Buffer.from(svg)
         });
-    } else {
-        file.base = path.resolve(file.base);
     }
+
+    file.base = path.resolve(file.base);
 
     // Add the shape to the internal processing queue
     this._queue.add(file);
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -117,15 +117,19 @@ SVGSpriter.prototype.add = function(file, name, svg) {
             throw e;
         }
 
+        // Resolve path before splitting it into base and path
+        // so that Shape will properly extract a shape name later on.
+        file = path.resolve(file);
+
         // Instantiate a vinyl file
         file = new File({
             base: path.dirname(file),
             path: file,
             contents: Buffer.from(svg)
         });
+    } else {
+        file.base = path.resolve(file.base);
     }
-
-    file.base = path.resolve(file.base);
 
     // Add the shape to the internal processing queue
     this._queue.add(file);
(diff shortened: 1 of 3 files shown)


XhmikosR:
Please add a test case.

Sidnioulz:
I've rebased and added a test case. I feel the way the test is written is clumsy, but I don't know how you prefer to have it organised. Either way, the test failed without the fix applied and passes with it applied.

test/svg-sprite.test.js
```diff
@@ -213,6 +220,30 @@ describe('svg-sprite', () => {
...
+
+        describe(`with ${weather.length} relative path SVG files`, () => {
+            const spriter = new SVGSpriter({
+                shape: {
+                    dest: 'svg'
+                }
+            });
+
+            it(`returns ${weather.length} optimized shapes`, function(done) {
+                this.timeout(20_000);
```
Kreeg, line 232: why do you add timeout?

XhmikosR: It could have been needed in the past, I removed it to see if CI is green

lib/svg-sprite.js
```diff
@@ -117,16 +117,20 @@ SVGSpriter.prototype.add = function(file, name, svg) {
...
+        // so that Shape will properly extract a shape name later on.
+        file = path.resolve(file);
+
         // Instantiate a vinyl file
         file = new File({
             base: path.dirname(file),
             path: file,
             contents: Buffer.from(svg)
         });
+    } else {
```
XhmikosR, line 130: This branch seems to be uncovered. Are we sure it's needed? Or are the tests enough?

Kreeg: It seems that the tests are enough for now

Sidnioulz:
Sorry, I didn't have time to go through this before the weekend. Thank you very much for taking the time to check the branch and tests!

lib/svg-sprite.js
XhmikosR: Nope, these are still uncovered.

Kreeg:
@Sidnioulz hello!
I have trouble with pushing commits to your fork's main branch.
Could you please rebase it from https://github.com/svg-sprite/svg-sprite/tree/fix-missing-shape-names ?

XhmikosR:
Branch rebased.

PR #378, opened 2021-01-04
Assorted fixes

### LINKED ISSUE #348: Remove lodash-pluck dependency
This is pretty old dependency and was replaced by `map`: https://github.com/lodash/lodash/wiki/Changelog#v400

XhmikosR:
@jkphl: unsure if the package is still maintained but since we are still using it, I thought I'd make a PR. I could split some of the changes if needed.

There are failing tests even on master on Node.js >=12. I think the test PNGs need to be updated and also:

* ~~svg2png; the current version breaks on Node.js 15~~ I updated it, but needs another pair of eyes
* ~~node-sass; the current version does not work on Node.js 15~~
* ~~mocha; still works though~~
* ~~istanbul should be replaced with nyc because it's deprecated~~ replace it with c8
* image-diff is deprecated and should be replaced too (could be one of the reasons image diffs fail on Node.js >= 12?)
* ~~pn can be removed and use Node.js's `util.promisify`~~
* updated svgo to v2.x, needs another pair of eyes
* JSHint needs to be replaced at some point too

~~The rest of the deps shouldn't matter but my guess is Node.js < 10 support will need to be dropped.~~ Edit, dropped it since more and more deps don't work there.

This should fix any [lgtm.com](https://lgtm.com/projects/g/jkphl/svg-sprite/?mode=list) issues. I also added a CodeQL action which is the successor to CodeQL, but ideally you could use both.


XhmikosR:
@jkphl friendly ping :)

I can split some of the patches; I started this for a few fixes but I've landed quite a few patches, so it might be easier for you if I split the branch somehow. That being said, I tried to keep the patches as clean as possible.

Please also check the PR description.

XhmikosR:
Closing this since I've split most patches. The only remaining patch is the svgo update which I'll tackle in the next days in a separate PR.

PR #398, opened 2021-02-23
Drop Node.js < 10 support and update most devDependencies

XhmikosR:
I need more eyes on this @jkphl due to the test files changes.

I also ended up adding a separate image for the failing test on Node.js > 10, but ideally, we should figure out why this happens and see if it's something in main or test code.

Also, note that I went with c8 for coverage since it's lighter than nyc, but I can switch to nyc if you prefer.

Split from #378 

TODO:

- [ ] Figure out why we are getting the output difference on Node.js >= 12.


XhmikosR:
It looks like the sorting algorithm has changed in Node.js >= 12 or in some other part of the code:

#### Node.js 10:

1. weather-storm~hover.svg
2. weather-clear.svg
3. weather-few-clouds.svg
4. weather-clear-night.svg
5. weather-severe-alert.svg
6. weather-few-clouds-night.svg
7. weather-overcast.svg
8. weather-showers.svg
9. weather-showers-scattered.svg
10. weather-snow.svg
11. weather-storm.svg

#### Node.js 12:

1. weather-storm~hover.svg
2. weather-clear.svg
3. weather-clear-night.svg
4. weather-overcast.svg
5. weather-storm.svg
6. weather-few-clouds.svg
7. weather-few-clouds-night.svg
8. weather-severe-alert.svg
9. weather-showers.svg
10. weather-showers-scattered.svg
11. weather-snow.svg

Basically this diff:

```diff
diff --git "a/C:\\Users\\xmr\\Desktop\\2.txt" "b/C:\\Users\\xmr\\Desktop\\1.txt"
index e27e0b9..8f4e7e1 100644
--- "a/C:\\Users\\xmr\\Desktop\\2.txt"
+++ "b/C:\\Users\\xmr\\Desktop\\1.txt"
@@ -1,11 +1,11 @@
 weather-storm~hover.svg
 weather-clear.svg
-weather-clear-night.svg
-weather-overcast.svg
-weather-storm.svg
 weather-few-clouds.svg
-weather-few-clouds-night.svg
+weather-clear-night.svg
 weather-severe-alert.svg
+weather-few-clouds-night.svg
+weather-overcast.svg
 weather-showers.svg
 weather-showers-scattered.svg
-weather-snow.svg
\ No newline at end of file
+weather-snow.svg
+weather-storm.svg
\ No newline at end of file

```

I still don't get where this change is coming from exactly.

XhmikosR:
OK, it might be something else, like mixing sync and async methods in test/svg-sprite?

XhmikosR:
OK, I think I pinpointed the issue: https://github.com/nodejs/node/issues/24294

jkphl:
@XhmikosR I didn't deep dive into the issue yet. But in general, the order of the images shouldn't really matter, as long as the result is consistent and we can create working test fixture for it. (It's possible that I didn't fully grasp the problem here yet. Sorry in that case …)

XhmikosR:
If you think the sorting code is fine, we can go with the solution I have here by using a different test image on Node.js > 10.

But the issue hints that something might be wrong in the sorting algo.

jkphl:
@XhmikosR I'm about checking this out now …

XhmikosR:
I think the problematic code is this https://github.com/svg-sprite/svg-sprite/blob/26ecb63a651ff09785cc30529467f91c858e30b7/lib/svg-sprite/mode/css/packer.js#L29-L31

We should return -1, 0, 1

jkphl:
@XhmikosR Alright, should be an easy one then, right? I'm still struggling with all the thousand tasks at the same time, so in case you're faster than me — all I'd do in this case would be something like:

```js
this.blocks.sort(function(a, b){ 
    const order = Math.max(b.width, b.height) - Math.max(a.width, a.height);
    return order ? (order / Math.abs(order)) : 0;
 }); 
```

Right?

XhmikosR:
I tried it but it doesn't seem to fix the issue; the output is still different on Node.js > 10. We'll figure it out now that we have pinpointed the root cause.

XhmikosR:
Actually @jkphl I think we can just go with the solution I have here. It's probably not worth spending time to get the same output on Node.js 10, although it'd be nice to.

I'll proceed with merging this tomorrow after adding a comment in tests.

jkphl:
@XhmikosR Alright! I'll wait until that merge then. I just figured I'm running Node v12 here on my (Windows) machine, so this issue is probably the main reason why many tests are failing for me right now. Let's pick it up from there then.

XhmikosR:
No, actually it's due to phantomjs and DPI. My dev VM is Windows 10 so I cannot run any tests properly. This will be fixed by puppeteer :)

Issue #365, opened 2020-09-02
Make "cssmin" optional for style blocks


Smidds:
### Some Context
When specifying animation keyframes to the `style` block, `cssmin` appears to destroy the definition of those animations!

**Example SVG**
```html
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 155 172">
    <defs>
        <style>
            @keyframes spin {
                0% {
                    transform:rotate(0deg);
                } 

                100% { 
                    transform:rotate(180deg);
                }
            }

            @keyframes dashHide { 
                0% {
                    clip-path: circle(35% at 50% 60%);
                } 

                100% {
                    clip-path: circle(75%);
                }
            }            
        </style>
    </defs>
</svg>
```

**Resulting `style` block after `cssmin`ification**
```css
'@keyframes spin{0%{transform:rotate(0deg)}to{transform:rotat{transform:rotate(180deg)}}@keyframes dashHide{0%{clip-path:circle(35%}@keyframes dashHide{0%{clip-path:circle(35% at 50% 60%)}to{clip-path:circle(75%)}}{clip-path:circle(75%)}}}'
```
It should, in this single-line form, look like this:
```css
@keyframes spin{0%{transform:rotate(0deg)}to{transform:rotate(180deg)}}@keyframes dashHide{0%{clip-path:circle(35% at 50% 60%)}to{clip-path:circle(75%)}}
```

### My Modest Proposla
Make the usage of `cssmin` optional! It seems `svgo` does a pretty good job cleaning up all my whitespace in this style block, and I'd like to remove this functionality seeing as I personally have no use for it and would like to preserve these animations.

I'm happy to make a pull-request adding an option passed via the config to disable, if that would be welcome!

Smidds:
Worth sharing, this is the problematic line in question: https://github.com/jkphl/svg-sprite/blob/master/lib/svg-sprite/shape.js#L728

XhmikosR:
An option wouldn't hurt, but IMHO cssmin should be replaced with something more maintained and better tested.

svgo itself is using ccso. Personally, I use clean-css but it would probably make more sense to use one of the dependencies that are in the deps tree.

XhmikosR:
I made #402, currently WIP, but csso is much better maintained and should fix your problem.

That being said, after the above lands, feel free to submit a PR making this configurable allowing enabling/disabling and/or passing custom options.

PR #402, opened 2021-02-23
Replace cssmin with csso

XhmikosR:
cssmin is unmaintained and probably has bugs.

I went with csso because svgo is already using it, so it's not another dependency. I also went with `restructure: false` to be safe.

It doesn't seem there are tests for CSS minification, from a quick look.

~~Depends on #398 and I'll probably need to rebase later.~~

Doesn't fi x #365 but shouldn't break that case anymore.

This is basically untested!

We need a test case for this, but we should at least test it manually.

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -24,6 +24,7 @@ execFile					= require('child_process').execFile,
 phantomjs					= require('phantomjs-prebuilt').path,
 dimensionsPhantomScript		= path.resolve(__dirname, 'shape/dimensions.phantom.js'),
 async						= require('async'),
+csso = require('csso'),
 /**
  * Default callback for shape ID generation
  *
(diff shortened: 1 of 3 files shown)


jkphl (approved):
I don't have experience with CSSO myself but code-wise and looking at the CSSO documentation this looks all correct 👍

XhmikosR:
@jkphl is this being tested? If not, could you please test it with an SVG with CSS?

jkphl:
@XhmikosR I can't tell whether the regular tests will work (as they don't work in general at the moment) but I did a quick manual / independent test with the Zebra image and that one at least looks good / worked.

XhmikosR:
Tests work on CI, but I don't know if csso is involved in any of those.

Do you have a quick cmd I could use to test this out too?

jkphl:
Use the `example.js` and change [line 5](https://github.com/svg-sprite/svg-sprite/blob/master/example.js#L5) to:

```js
cwd = path.join(__dirname, 'test', 'fixture', 'svg', 'css'),
```

and [line 7](https://github.com/svg-sprite/svg-sprite/blob/master/example.js#L7) to

```js
files = glob.sync('**/zebra.svg', { cwd: cwd });
```

then run `node example.js`. This should give you an SVG in `tmp/css/svg` containing minified CSS rules with replaced IDs (that's the letter "a" in front of all IDs), so that's exactly the output of the altered lines. They look good to me.

XhmikosR:
@jkphl thanks for the script. That being said that file also works on master AFAICT...

XhmikosR:
So, I'm still not sure if this works :P

jkphl:
@XhmikosR Well, the most important thing is that it doesn't **break** anything either, right? 😉

XhmikosR:
Well, if it's uncovered by the tests, it can't break :P

That being said, the docs hint my patch is fine and tests do pass anyway. I'll merge it and we see how it goes.

jkphl:
Well, it kinda is covered (or at least was at some point), because otherwise the visual output tests would fail. However, we can't tell this until the tests run fine again in general. Until then, only a manual check will help — which seems to look good so far. Or do I miss something?

XhmikosR:
We should be fine :)

PR #403, opened 2021-02-24
Update svgo to v2.3.0

XhmikosR:
~~Requires #398 and I'll need to rebase it.~~

This changes the output sprites because some defaults have changed and some bugs are fixed upstream. We could change the svgo options to match the previous behavior, but I'm not sure if we should. Namely `convertPathData`'s `noSpaceAfterFlags` option is now false: https://github.com/svg/svgo/pull/1353

Personally, I think it's better to stick to the defaults + only the plugins we need to change which I assume is only setting `inlineStyles` to `false`, although I'm not sure why. If it's only so that the tests are passing, I have [reported](https://github.com/svg/svgo/issues/1359) the issue upstream regarding zebra.svg and I guess we could replace it with a simpler file in the meantime.

TODO:

- [x] Update test files
- [x] Make sure custom config still works
- [x] Verify the expected file changes are all good
- [x] Decide if we still need to change any svgo default options or not


lib/svg-sprite/transform/svgo.js
```diff
@@ -70,15 +30,20 @@ var SVGO = require('svgo'),
...
+    });
 
     var svg = shape.getSVG(false),
-        svgLength = svg.length,
-        svgoInstance = new SVGO(config);
+        svgLength = svg.length;
 
-    svgoInstance.optimize(svg).then(function (result) {
+    try {
+        var result = svgo.optimize(svg, config);
```
XhmikosR, line 46: Need another pair of eyes but this should be synchronous in v2.x.

jkphl: Yeah, you should be right! Looks good to me.

jkphl (approved):
I agree that we should stick to the defaults as much as possible, so this looks fine for me. Listing (and commenting out) all the SVGO options here was just a sort of convenience thing, as many issues people reported were like "I have this beautiful SVG here and it gets destroyed when it's added to the sprite". In 99% of the cases it was an overly aggressive (default) SVGO plugin which was the reason for this. So I tried to ease things a bit by adding the settings here and pointing people to it when they experienced problems. It's no must, and we could re-add them later if we want. On the other hand, it's a tedious job to keep them in sync with the defaults … In any case, let's move on to 2.x!

XhmikosR:
@jkphl Then, I think we should stick the svgo defaults. I'll keep this on hold, perhaps we need to move with #321?

XhmikosR:
@jkphl I went ahead and kept the default svgo options. I think it makes more sense since it's not our job to work around any svgo issues. Instead, any issues should be reported upstream so that they are fixed and everyone benefits.

Do note that with the default svgo options I notice a 1px change in some images.

PS. I kept the patches separate, but I'll probably squash into on when merging.

lib/svg-sprite/transform/svgo.js
```diff
@@ -70,15 +24,20 @@ var SVGO = require('svgo'),
  */
 module.exports = function (shape, config, spriter, cb) {
     config = _.cloneDeep(config);
-    config.plugins = ('plugins' in config) ? defaultPluginConfig.concat(config.plugins) : defaultPluginConfig;
-    config.plugins.push({ removeXMLProcInst: !!spriter.config.svg.xmlDeclaration });
-    config.plugins.push({ removeDoctype: !!spriter.config.svg.doctypeDeclaration });
+    config.plugins = 'plugins' in config ? svgo.extendDefaultPlugins([...config.plugins]) : [];
```
XhmikosR, line 27: Would be nice to confirm if custom SVGO options still work; we don't have a test for this.

lib/svg-sprite/transform/svgo.js
```diff
@@ -70,15 +24,22 @@ var SVGO = require('svgo'),
  */
 module.exports = function (shape, config, spriter, cb) {
     config = _.cloneDeep(config);
-    config.plugins = ('plugins' in config) ? defaultPluginConfig.concat(config.plugins) : defaultPluginConfig;
-    config.plugins.push({ removeXMLProcInst: !!spriter.config.svg.xmlDeclaration });
-    config.plugins.push({ removeDoctype: !!spriter.config.svg.doctypeDeclaration });
+    config.plugins = 'plugins' in config ? [...config.plugins] : [];
```
XhmikosR, line 27: Would be nice to confirm if custom SVGO options still work; we don't have a test for this.

lib/svg-sprite/transform/svgo.js
```diff
@@ -70,15 +24,22 @@ var SVGO = require('svgo'),
...
+    config.plugins.push({
+        name: 'removeXMLProcInst',
+        active: !!spriter.config.svg.xmlDeclaration
+    }, {
+        name: 'removeDoctype',
+        active: !!spriter.config.svg.doctypeDeclaration
+    });
+
+   config.plugins = svgo.extendDefaultPlugins([...config.plugins]);
```
XhmikosR, line 36: Need another pair of eyes here in case spread isn't needed.

XhmikosR:
Some more insight from svgo's 2.x [docs](https://github.com/XhmikosR/svgo#configuration):

>The default list is fully overridden if the `plugins` field is specified. To extend the default list use the `extendDefaultPlugins` utility:
>
>```js
>const { extendDefaultPlugins } = require('svgo');
>module.exports = {
>  plugins: extendDefaultPlugins([
>    {
>      name: 'builtinPluginName',
>      params: {
>        optionName: 'optionValue'
>      }
>    }
>  ])
>}
```

File: lib/svg-sprite/transform/svgo.js (hunk 3 above)
Comment (jkphl) [reply lib/svg-sprite/transform/svgo.js:27]: I'm not too savvy with the spread syntax in JavaScript and might overlook some (wanted) side effects (like implicit type casting?), but wouldn't the following do the same here (now that we drop `defaultPluginConfig`):

```js
config.plugins = config.plugins || []; // assuming config.plugins always is an array
```

or

```
if (!config.hasOwnProperty('plugins')) {
    config.plugins = [];
}
```

jkphl (approved):
Assuming the binary files in this PR match the new SVGO version (I didn't check them in detail) this one looks good to me!

lib/svg-sprite/transform/svgo.js
XhmikosR: Hmm, indeed, the spread there might be redundant. It's a pity we don't have tests, so we need to test this out somehow manually for now.

XhmikosR:
BTW this branch results in a quite bigger SVG. For the example we have in the repo 158.8 kB -> 262.1 kB. Something is off it seems.

I'll try to figure this out tomorrow before merging this because it's a serious issue.

XhmikosR:
After talking with the svgo main developer, it seems this is expected right now. It will improve in the feature, but the defaults are used so that the output is correct, not the smallest one.

PR #417, opened 2021-03-01
Replace image-diff with looks-same

XhmikosR:
Split from #284 

TODO:

- [ ] Switch to promises
- [ ] Write the diff image only if an error occurs?


XhmikosR:
@mojoaxel LGTY?

XhmikosR:
I'm going to merge this for now and we improve it later.

PR #418, opened 2021-03-01
Replace `_.trim` with native JS

XhmikosR:
Second attempt after #408.

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -84,15 +84,15 @@ SVGSpriter.prototype._namespacePow = [];
 SVGSpriter.prototype.add = function(file, name, svg) {
     // If no vinyl file object has been given
     if (!this._isVinylFile(file)) {
-        file = _.trim(file);
+        file = file.trim();
         let error = null;
 
         // If the name part of the file path is absolute
         if (name && path.isAbsolute(name)) {
             error = format('SVGSpriter.add: "%s" is not a valid relative file name', name);
         } else {
-            name = _.trimStart(_.trim(name), `${path.sep}.`) || path.basename(file);
-            svg = _.trim(svg);
+            name = _.trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
+            svg = typeof svg === 'string' ? svg.trim() : svg;
 
             // Argument validation
             if (arguments.length < 3) {


lib/svg-sprite.js
```diff
@@ -96,8 +96,8 @@ SVGSpriter.prototype.add = function (file, name, svg) {
 
             // Else
         } else {
-            name = _.trimStart(_.trim(name), path.sep + '.') || path.basename(file);
-            svg = _.trim(svg);
+            name = _.trimStart(name.trim(), path.sep + '.') || path.basename(file);
+            svg = typeof svg === 'string' ? svg.trim() : svg;
```
XhmikosR, line 100: @jkphl So, the question is do we need to trim here? `svg` can be a Buffer which resulted in failures without this check, but it makes me wonder.

jkphl: @XhmikosR Hmmm … I'm a bit unsure here (and cannot test in this very moment). It's possible that `_.trim()` has the neat side effect to automatically serialize Buffers and convert non-strings to empty strings, so that `svg` reliably is a string in the end. Can you confirm this? If so, we should ensure that `svg` ends up as a string in the end. Apart from that, we can surely use the native method.

XhmikosR: That I don't know TBH. It wasn't caught by the tests at all, I just hit it since I'm testing the master and lint branches on https://github.com/twbs/icons/blob/main/package.json#L28

XhmikosR: @jkphl yup, _.trim does cast the value to string before trimming.

So, the question is if we need to convert this to a string if it's not one? In my case svg is a Buffer and it's only used a few lines below for Vinyl contents. Which further makes me wonder if `Buffer.from` is needed there in case svg is already a Buffer?

Anyway, we can clean that up later, separately from this.

jkphl: I'm not absolutely sure about my original intentions here. I guess I wanted to allow both string and Buffer arguments (or possibly I didn't expect Buffer arguments at all in the beginning and they just happened to work?). Using `_.trim()` on the `svg` parameter [makes sure we don't add](https://github.com/svg-sprite/svg-sprite/blob/master/lib/svg-sprite.js#L112) invalid shapes consisting of, say, whitespace characters only. Not sure how easily one can ensure this for a Buffer as well. Casting a Buffer to a string and then re-creating as a Buffer definitely isn't necessary. But we would have to make sure that also string arguments still work as we don't know how people use the `add()` method. So maybe you're right and we should leave this as is for the time being and come back at a later point.

XhmikosR: There are also `Buffer.isBuffer()` or `vinyl.isBuffer()`. Either way, this is not currently well covered by the tests, so let's leave it for now and we'll revisit later.

Kreeg:
@XhmikosR I think we can go back to this later.

PR #419, opened 2021-03-01
Update js-yaml to v4.0.0

XhmikosR:
@jkphl need to confirm we are not affected by the breaking changes apart from `safeLoad` -> `load`: https://github.com/nodeca/js-yaml/blob/master/migrate_v3_to_v4.md

We need to check the yaml files.


jkphl (approved):
The YAML files used for aligning and annotating are super trivial, and even though the test in general don't work at the moment, changing the alignment YAML does work as expected, so I guess we can assume this is safe!

PR #423, opened 2021-03-05
Replace `_.map` with vanilla JS

XhmikosR:


jkphl (approved):
Looks a bit more verbose, but should be fine 👍

PR #425, opened 2021-03-05
Replace `_.reject` with native JS

XhmikosR:


jkphl (approved):
Of course — no-brainers!

PR #426, opened 2021-03-06
Use the path methods when possible

XhmikosR:


jkphl (approved):
Ok, these ones baffle me a bit, to be honest. Not using `path.x()` here seems just stupid. So I'm wondering whether I was just stupid when I wrote them 😉 — or if I had some sort of reason. Anyway, as long as the tests succeed, it should be safe to use the path ones. (2 of the occurrences are log messages only, anyway, so definitely no harm done …)

XhmikosR:
TBH, I wondered know why you didn't use the path methods either :/

I believe it makes sense, but we might be missing something and tests don't cover everything. That being said, I feel like this is the proper way and worst case we revert it and document it with a comment.

jkphl:
💯

XhmikosR:
One thing that comes to mind is maybe you wanted to enforce `/` even on Windows? If so, there's [`path.posix`](https://nodejs.org/docs/latest-v10.x/api/path.html#path_path_posix) we could use. After we move to puppeteer I'll add Windows tests too since phantomjs is broken on Windows.

jkphl:
@XhmikosR I really can't tell if that was my reason, but it's possible. I've always been working on Windows workstations while our servers are Linux only, so basically it always needed to run on both ...

PR #427, opened 2021-03-06
layouter.js: cache length variable before the loop

XhmikosR:
Not sure if this makes any difference performance-wise in this specific case.


jkphl (approved):
Yep, sure! You *could* consider making this a `const` already, while you are at it, but we could tackle a general overhaul separately as well, of course.

XhmikosR:
ES6-ifying is coming after these PRs https://github.com/svg-sprite/svg-sprite/compare/lint?w=1 :)

PR #428, opened 2021-03-07
Replace indexOf with startsWith/endsWith

XhmikosR:
So, this is definitely better IMHO as long as `string` isn't undefined.

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -682,7 +682,7 @@ SVGShape.prototype.setNamespace = function(ns) {
     		// Substitute ID references in xlink:href attributes
     		select('//@xlink:href', this.dom).forEach(function(xlink){
     			var xlinkValue			= xlink.nodeValue;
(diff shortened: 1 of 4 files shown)


lib/svg-sprite/shape.js
```diff
@@ -682,7 +682,7 @@ SVGShape.prototype.setNamespace = function(ns) {
     		// Substitute ID references in xlink:href attributes
     		select('//@xlink:href', this.dom).forEach(function(xlink){
     			var xlinkValue			= xlink.nodeValue;
-    			if ((xlinkValue.indexOf('data:') !== 0) && (xlinkValue in substIds)) {
+    			if (!xlinkValue.startsWith('data:') && (xlinkValue in substIds)) {
```
XhmikosR, line 685: I need a confirmation for this case

jkphl: Looks good for me!

jkphl (approved):
They should all do their job I suppose 👍

PR #431, opened 2021-03-09
Fix lodash cloneDeep usage

XhmikosR:
@jkphl now this might hide some issues that went unnoticed until now.

It seems that the `isDeep` parameter was removed in lodash v4.0.0. There's the `cloneDeep` function, but I don't know if it's supposed to be used in these cases.

EDIT: we should really have tests for this. :/ Feel free to push a test.

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -209,7 +209,7 @@ SVGSpriter.prototype._transformShape = function (shape, cb) {
 SVGSpriter.prototype.compile = function(config, cb) {
     var _config = _.isPlainObject(config) ?
         this.config.filter(config) :
-        _.clone(this.config.mode, true);
+        _.cloneDeep(this.config.mode);
     var _cb;
 
     if (_.isFunction(cb)) {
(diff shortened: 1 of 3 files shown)


XhmikosR:
@jkphl any thoughts about this? As things are, the parameter is moot. What I don't know is if you meant to use `cloneDeep` in any of these places.

jkphl:
@XhmikosR From what I can tell (and read from the current lodash docs), `cloneDeep` instead of `clone` is what we need in all of these cases, and it should be completely harmless to swap the two (`cloneDeep` will be slightly more resource demanding but that doesn't really count in our case). It's just important to really decouple the clone from the original in order to not unintentionally alter values in there, and a shallow clone would keep references to the original … I'd say it's safe to use `cloneDeep` here, so this is what we should do.

XhmikosR:
@jkphl I see, just keep in mind that for years now (since you updated to lodash 4.x) you are basically using `clone`. I don't have the time to investigate this more, though.

jkphl:
@XhmikosR You're totally right — it wouldn't surprise me if this has already troubled some folks in the past. That said, using `cloneDeep` definitely makes things safer. Fixing this might not be necessary, but now that you became aware of the deprecation … Let's just do this!

jkphl (approved):

PR #436, opened 2021-03-12
Enforce Lint and modernize codebase

XhmikosR:
@jkphl I open this to land it after a thorough review. Remember to always add `?w=1` after each patch to view the non-whitespace diff.

I need a second pair of eyes for some of the patches I kept separate, but also check everything out please so that we are safe :)

Non-whitespace diff: https://github.com/svg-sprite/svg-sprite/pull/436/files?w=1

TODO:

- [x] Review thoroughly each patch
- [x] See if we can address any newly added TODO comments
- [x] Enable these currently disabled rules:
  - [x] `no-labels` -> **I need help with this**
- [ ] Clear up a few situations with forEach thisArg binding: -> **I need help with this**
  - [ ] https://github.com/svg-sprite/svg-sprite/blob/eefe61306d1896d8703d35c498d069c79172ffbb/bin/svg-sprite.js#L225-L238
  - [ ] https://github.com/svg-sprite/svg-sprite/blob/23377c0cd8bcc40fcbb0fb2e318b0922ff105eb7/bin/svg-sprite.js#L242-L259
  - [ ] https://github.com/svg-sprite/svg-sprite/blob/23377c0cd8bcc40fcbb0fb2e318b0922ff105eb7/lib/svg-sprite/config.js#L197-L222
- [ ] Rebase/squash any patches before the final merge


bin/svg-sprite.js
```diff
@@ -44,39 +45,38 @@ var _ = require('lodash'),
...
+        } else if (option.required) {
+            yargs = yargs.require(alias);
+        }
+
+        if ('map' in option) {
+            optionsMap[option.map] = name;
+        }
+    }
+
+    const { description, alias: optAlias, default: optDefault, map, ...children } = option;
```
jkphl, line 76: Im not 1000% sure if I already fully understand what's going on here but I somehow suspect it should be `optionsMap` here instead of `map`. Possible?

XhmikosR: `map` is the `option.map` key we are ignoring like everything else before that too. We only want the children.

jkphl: Got it, thanks. I was just wondering because there's another `const map` down below in line 138. But maybe I can't fully see the scopes here in the diff view ...

XhmikosR: Hence why I suggesting viewing each patch separately with the `?w=1` query appended :)

jkphl: I'm sorry I'm really unexperienced with this. I'm using `?w=1`, so whitespace is not an issue. However, I'm rather viewing file by file, which might then be several patches as once, right? Where can I find links for patch-wise changes? Thanks for bearing with me!

XhmikosR: https://github.com/svg-sprite/svg-sprite/pull/436/commits and open each patch to a new tab. There's also a cog button to hide whitespace there :)

jkphl: Thanks! So it's commit-wise — I thought there was a difference to patches. 😉 I'll have to take a break now (and do what my clients expect me to do on a Sunday 😥) but I'll come back later today / tonight!

jkphl (approved):

jkphl:
@XhmikosR Oh, it looks like I had approved the whole PR although I only wanted to approve one commit (more as an internal note to myself). Please disregard the approval 3 minutes ago then!

XhmikosR:
@jkphl I cleaned up the branch and moved some of the patches to another branch. I'll split those later after this one lands.

lib/svg-sprite.js
```diff
@@ -166,19 +166,18 @@ SVGSpriter.prototype._isVinylFile = function(file) {
  */
 SVGSpriter.prototype._transformShape = function(shape, cb) {
     const tasks = [];
-    const that = this;
-    const createTransformTask = function(transform) {
+    const createTransformTask = transform => {
         // If it's a custom transformer
         if (_.isFunction(transform[1])) {
```
XhmikosR, line 171: Ideally, I'd like to destructure transform[0] and transform[1] too, but maybe we leave this for later.

lib/svg-sprite.js
```diff
@@ -249,40 +248,33 @@ SVGSpriter.prototype._compile = function () {
...
                 // Add intermediate SVG files
                 if (that.config.shape.dest) {
                     files.shapes = that._getShapeFiles(that.config.shape.dest);
                     that.verbose('Returning %d intermediate SVG files', files.shapes.length);
                 }
-                that.info('Finished %s sprite compilation', Object.keys(data).map(function (mode) {
-                    return '«' + mode + '»';
-                }).join(' + '));
+
+                that.info('Finished %s sprite compilation', Object.keys(data).map(mode => '«' + mode + '»').join(' + '));
```
jkphl: We could probably replace `'«' + mode + '»'` with a template string, right? Minor thing though …

XhmikosR: Yup, I played with the idea of enforcing template literals, but they barely looks more readable in only a few cases. So, I decided to leave it out for the time being.

XhmikosR: See https://github.com/svg-sprite/svg-sprite/commit/2f66c4fa529d49e1f70d73f8b591ce2ed494e76d

I can pick this patch in this branch later if you want, otherwise I'll just land it separately.

jkphl: Whatever works best for you!

jkphl (commented):
Totally not against these changes, but I'm just curious: What's the motivation behind these changes? From my hardcore performance optimization experience (in the PHP world, though, and that was many, many years ago so it might not be even valid knowledge anymore) I'm trained to

* use pre-increments instead of post-increments wherever possible as for some mysterious reason they used to be way more performant,
* avoid as many costly checks as possible in loop conditions (although counting array lengths in JavaScript might be more efficient than in PHP).

Well, as I said, I'm just curious (and I'm definitely carrying over experience from one programming language to another so that might be a problem itself 😉).

jkphl (commented):
Similar [here](https://github.com/svg-sprite/svg-sprite/pull/436/commits/8693df3d3c1fbffeebeda1c2c117b6176eb7ca91): I'm totally fine with removing the commented out parameters. Just as an explanation: I usually keep those as comments to a) make linting happy but b) still inform those looking at the code that these arguments would be available if needed. Sort of service-tweets for developers. 😉

XhmikosR:
> Totally not against [these changes](https://github.com/svg-sprite/svg-sprite/pull/436/commits/6ee6fbec2ce1731a403aaf5c0188efddd98afd2e), but I'm just curious: What's the motivation behind these changes? From my hardcore performance optimization experience (in the PHP world, though, and that was many, many years ago so it might not be even valid knowledge anymore) I'm trained to
> 
>     * use pre-increments instead of post-increments wherever possible as for some mysterious reason they used to be way more performant,
> 
>     * avoid as many costly checks as possible in loop conditions (although counting array lengths in JavaScript might be more efficient than in PHP).
> 
> 
> Well, as I said, I'm just curious (and I'm definitely carrying over experience from one programming language to another so that might be a problem itself 😉).

Well, post/pre increments don't seem to make any difference from my benchmarks, and in fact post increments are faster. As for the cached length variable, it also doesn't seem to make any important difference either. Some years ago, the length check did certainly matter, but nowadays doesn't seem to matter :)

Node.js and JS engines are constantly improving.

> 
> 
> Similar [here](https://github.com/svg-sprite/svg-sprite/pull/436/commits/8693df3d3c1fbffeebeda1c2c117b6176eb7ca91): I'm totally fine with removing the commented out parameters. Just as an explanation: I usually keep those as comments to a) make linting happy but b) still inform those looking at the code that these arguments would be available if needed. Sort of service-tweets for developers. 😉

a) we don't need it with linting AFAICT?
b) how would they be available? The reason I removed them is because they aren't available?

jkphl (approved):
@XhmikosR … which means: I'm finally done with reviewing all these patches, and from what I can see they look just perfect! I didn't spot anything suspicious. AWESOME JOB! 🎉

XhmikosR:
@jkphl Great! I still need help with the TODO I have in the PR description. Ideally, we should get those sorted out here.

jkphl:
> Well, post/pre increments don't seem to make any difference from my benchmarks, and in fact post increments are faster. As for the cached length variable, it also doesn't seem to make any important difference either. Some years ago, the length check did certainly matter, but nowadays doesn't seem to matter :)

That's interesting insights, thank you! And yes, this is also what you would expect after all these years (I did my last benchmarks — in the PHP world! — many years ago …). Also, these tiny benefits — if any — mostly weigh in when used thousands or millions of times, which definitely isn't the case here. But still, interesting knowledge — thanks!
 
> a) we don't need it with linting AFAICT?
> b) how would they be available? The reason I removed them is because they aren't available?

So it was you who commented them out earlier? It just looked a lot like me, I'm doing that all the time (and usually I'm doing this only when such arguments truly exist but linting would yell at me as they aren't used anywhere). However, if the really don't exist anymore, they should definitely be removed. As I said, just go ahead with this — I just wanted to explain where the comments came from in the first place.

jkphl:
> @jkphl Great! I still need help with the TODO I have in the PR description. Ideally, we should get those sorted out here.

Well, let's try …

> `no-labels` -> I need help with this

Is this something regarding the use of xo for linting? TBH, I have never heard of xo before, so I'm not familiar with this rule yet. Are there anymore details about problematic parts in the code?

> Clear up a few situations with forEach thisArg binding

Regarding the examples you mentioned: Not wanting to be blasphemous, but is there anything wrong with using `forEach()` and binding it to a context? All three examples seem legit (although the third one does look a bit over-engineered as it's only one item that's iterated over …). I'd just like to understand the need for change before digging further into it.

XhmikosR:
Regarding labels, I find the code harder to follow which is the reasoning
behind this ESLint rule.

Regarding this argument, the reasoning is the same: clarity. Traditionally
this has been a big issue in JavaScript so TBH I just try to keep things
easy to grasp. There's no rule for this, but if you noticed I had to
manually get rid of this where arrow functions are used.

So, I believe we should clear this up if possible.

I'm not enabling here what xo has, but rather what I agree with :)

XhmikosR:
I just replied, I'm also AFK.

Otherwise I'll ping you again in the next days :)

jkphl:
> Regarding labels, I find the code harder to follow which is the reasoning behind this ESLint rule.

I somehow think I didn't fully understand what you're referring to with `no-labels`. Whenever you've got a minute to spend, it would be great if you could point me to some more information and / or locations in the code. 🙏

> Regarding this argument, the reasoning is the same: clarity. Traditionally this has been a big issue in JavaScript so TBH I just try to keep things easy to grasp. There's no rule for this, but if you noticed I had to manually get rid of this where arrow functions are used. So, I believe we should clear this up if possible.

I see. Well, we can definitely re-write the code here, possibly introducing new methods for the `forEach` callbacks. I'm not totally convinced that the results will be really easier to grasp, but I get your point. I might be able to make an attempt in the next few days — please bear with me here.

XhmikosR:
There's just one place with labels, if you check our the branch locally and
remove the rule suppression you should see it :)

I'm all for landing this with the label sorted out and we can leave this
arg for later. I just want to move forward :)

jkphl:
> There's just one place with labels, if you check our the branch locally and remove the rule suppression you should see it :)

Great, thanks! I'll do that as soon as I find some more time. 👍

lib/svg-sprite/layouter.js
```diff
@@ -10,116 +10,115 @@
...
 
     // Register the common shapes data
-    var shapesLength = this._spriter._shapes.length;
-    this._spriter._shapes.forEach(function(shape, index) {
-        var dimensions              = shape.getDimensions(),
-        padding                     = shape.config.spacing.padding;
+    const shapesLength = this._spriter._shapes.length;
+    this._spriter._shapes.forEach((shape, index) => {
+        const { width, height } = shape.getDimensions();
+        const { top, right, bottom, left } = shape.config.spacing.padding;
```
XhmikosR, line 104: @jkphl I think this might not be 100% right. Isn't `padding` an Array? If so, shouldn't we use this (untested)?

```js
const [top, right, bottom, left] = shape.config.spacing.padding;
```

There might be more places like this.

XhmikosR: OK, NVM, it seems it's an object...

PR #438, opened 2021-03-13
packer.js: stop nesting ternary expressions

XhmikosR:
@jkphl this is a tricky one, so bear with me please :)

I attempted to make this clearer because previously it was impossible to follow the logic. I'm not 100% sure if our tests cover this so please review thoroughly.

Split from #436


jkphl (approved):
One ternary from hell … well, looks perfect! 😉

XhmikosR:
I rely on your judgement because I honestly can't even recall how I unwrapped this shit :P

PR #439, opened 2021-03-13
base.js: use string.endsWith

XhmikosR:
@jkphl unsure why you were ignoring the case here, but I think this solution should be as close as the previous code, but clearer.

That being said, if `renderConfig.dest` can be undefined/null or not a string, this will throw now

### DIFF
--- a/lib/svg-sprite/mode/base.js
+++ b/lib/svg-sprite/mode/base.js
@@ -58,7 +58,7 @@ function SVGSpriteBase(spriter, config, data, key) {
 				}
 				if ('dest' in this.config.render[extension]) {
 					renderConfig.dest			= path.resolve(this.config.dest, this.config.render[extension].dest);
-					if (!renderConfig.dest.match(new RegExp('\\.' + extension + '$', 'i'))) {
+					if (!renderConfig.dest.endsWith('.' + extension.toLowerCase())) {
 						renderConfig.dest		+= '.' + extension;
 					}
 				}


jkphl (commented):
In general, using `endsWith()` here makes total sense for me (I guess it didn't exist or wasn't really popular when I wrote that). However, as `endsWith()` is case-sensitive we should make sure that both sides of the comparison are normalized. What about altering to `renderConfig.dest.toLowerCase().endsWith('.' + extension.toLowerCase())`?

XhmikosR:
Not sure I follow why we need to ignore case in the first place? Regardless, unless I'm mistaken, this patch should be equivalent with the previous solution. What you suggest would be new behavior?

jkphl:
@XhmikosR I might be missing something, so please correct me in that case, but I think there's a subtle difference between the patch and the previous behaviour:

* Before, the config value `renderConfig.dest` (as given by the user, could be lower case or upper case or mixed) was matched case-insensitively against a potential file extension, and the extension was appended in case it wasn't already part of the path.
* With the patch, the `renderConfig.dest` is compared to the lower-cased version of the file extension in a case-sensitive way (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith#description) at least). So `some/path.CSS` (= configured by the user) wouldn't match `.css`, resulting in `some/path.CSS.css`.

```js
if (!renderConfig.dest.toLowerCase().endsWith('.' + extension.toLowerCase())) {
    // …
}
```

should fix this, right?

XhmikosR:
I know that `endsWith` ignores case, but the behavior is the same I think. Previously, the regex compared case insensitively the extension, which is the same as this patch, isn't it? You weren't checking render.dest but only the last part, which is the extension.

jkphl:
@XhmikosR I was relying on MDN saying that `endsWith()` is case-sensitive (without testing this myself). If it's case-insensitive, however, and we can rely on this, then YES, the patch should be fine!

XhmikosR:
`endsWith` is case sensitive, hence why I'm lowercasing the extension. Sorry, that was a typo above.

jkphl:
@XhmikosR But wouldn't you have to make sure the user-defined config value `renderConfig.dest` is lower-cased as well before comparing the two? I should add that *for the result* it doesn't matter whether the extension is lower or upper case. Instead it should follow the user's wish. But *for the comparison* we need to make sure that both sides of the equation are normalized, shouldn't we? Hence the `renderConfig.dest.toLowerCase()` just for comparison purposes. What am I missing? 🤔

XhmikosR:
Either I'm too tired or you are too :P

So, previously, `renderConfig.dest` wasn't lowercased either AFAICT. Only the extension part was, right? Which is what I'm doing in this patch too.

jkphl:
> So, previously, renderConfig.dest wasn't lowercased either AFAICT.

Yeah, but that's why I was using a case-insensitive regex. Right? 😉

XhmikosR:
But I'm lowercasing the extension part which is what the regex also checked?

XhmikosR:
Either way, let's leave this for now and move to another PR. This is a trivial patch, we can revisit later.

jkphl:
Ok, let me try some examples. **Previously**, the lines in question were:

```js
if (!renderConfig.dest.match(new RegExp('\\.' + extension + '$', 'i'))) {
	// …
}
```

Assuming `extension` was `"css"` (or `"CSS"`, doesn't really matter), and ignoring the negation in the condition, the test would yield:

* `renderConfig.dest = "path/to/something.css"` → `true`
* `renderConfig.dest = "path/to/something.CSS"` → `true`
* `renderConfig.dest = "path/to/something.cSs"` → `true`
* `renderConfig.dest = "path/to/something"` → `false`

With the **new code**

```js
if (!renderConfig.dest.endsWith('.' + extension.toLowerCase())) {
	// …
}
```

we'd always compare to the lower-cased extension, i.e. `"css"`, in a case-sensitive manner, which would yield:

* `renderConfig.dest = "path/to/something.css"` → `true`
* `renderConfig.dest = "path/to/something.CSS"` → `false`
* `renderConfig.dest = "path/to/something.cSs"` → `false`
* `renderConfig.dest = "path/to/something"` → `false`

Please tell me if I made a mistake, but I see quite some differences in the results?

XhmikosR:
No, you are right. My mind was stuck at believing that lowercase === ignoring case. Let's drop this for now and I will revisit later if needed.

Thanks for pointing out the obvious, I need to sleep more :)

jkphl:
@XhmikosR No worries! I very much appreciate your efforts, so I'm just trying to look at least somewhat eager. 😉

PR #440, opened 2021-03-13
Make `SVGSpriter.compile()`'s arguments explicit

XhmikosR:
@jkphl bear with me here, I'm trying to be explicit and clearer. Instead of using `arguments` and shifting, the function accepts two parameters, which matches the jsDoc comments too.

Generally, optional parameters are at the end, though, but I tried to keep the previous behavior.

Split from #436

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -206,15 +206,24 @@ SVGSpriter.prototype._transformShape = function (shape, cb) {
  * @param {Object} config                Configuration
  * @param {Function} cb                    Callback
  */
-SVGSpriter.prototype.compile = function () {
-    var args = _.toArray(arguments),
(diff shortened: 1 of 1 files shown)


jkphl (approved):
This looks good to me. I didn't run any tests in particular, but it's so central it should be covered by the general tests, I suppose.

About the optionality of arguments: I guess I wanted to mimic jQuery's typical behaviour here, accepting various mixtures of arguments, identifiable by their type. Being more explicit would probably be my choice nowadays as well. :)

XhmikosR:
I just want to make sure I didn't f up the logic there, because I have to say there are some parts in the codebase like magic :P

jkphl:
@XhmikosR Yeah, well … I was young(er) and that was the spirit. 😉

XhmikosR:
As long as you find the logic the same, we can proceed. It's just that right now, I can't even follow the logic :P

PR #441, opened 2021-03-13
shape.js: simplify ENTITY regex

XhmikosR:
Split from #436.

@jkphl XML is case sensitive, isn't it? If so, the regex shouldn't use the `i` flag, right?

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -330,7 +330,7 @@ SVGShape.prototype._initSVG = function() {
 	}
 
 	// Resolve XML entities
-	var entityRegExp		= /<!ENTITY\s+(\S+)\s+(["'])(.+)\2>;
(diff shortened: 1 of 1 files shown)


jkphl (approved):

XhmikosR:
@jkphl shouldn't we remove the ignore case flag?

jkphl:
@XhmikosR We could probably do this, yet. But it's not particularly harmful either, right?

XhmikosR:
Not harmful, but IIRC it's not valid XML so we shouldn't care about it?

jkphl:
I'm fine with its removal. Just wanted to drop that I usually try to stick to [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle):

> Be conservative in what you do, be liberal in what you accept from others

😉

XhmikosR:
While I agree to some extent, I don't think we shouldn't rely on the specs. :)

I'll drop the ignore flag here too.

PR #442, opened 2021-03-13
bin/svg-sprite.js: use `_.merge` instead of the custom function

XhmikosR:
Makes more sense to be consistent for the time.

Unfortunately, the CLI isn't tested, so let me know if everything still looks good @jkphl

### DIFF
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -99,28 +99,6 @@ function addConfigMap(store, path, value) {
 }
 
 /**
- * Recursively merge two config objects
- *
- * @param {Object} from Source configuration
- * @param {Object} to Target configuration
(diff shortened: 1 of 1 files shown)


jkphl (approved):
This is a tricky one, as like you said the merge of two objects itself isn't tested, unfortunately. I can't guarantee that I don't miss any edge case but I performed some manual tests by using a custom config file and comparing the merged results before and after the change. It looked good, at least with real life config variables. I think we can risk changing this and then deal with problems in case any arise.

XhmikosR:
My plan is to reduce the lodash stuff we use and when we have them to a minimum see what else we could use. So, this makes sense because at least we are using something tested upstream now.

PR #444, opened 2021-03-17
chore: use .nvmrc

ert78gb:
I like to use `.nvmrc` file. It helps to identify which Node.js version recommended for the development. Enough just run `nvm use` or `nvm install` command.

We can put only the major version into the `.nvmrc` file, but few times node changes the `npm` version that generate different `package-lock.json` file.


XhmikosR:
This has no effect on Windows, though. And also, tying down the version so much isn't good IMHO.

The problem you are having is because of npm 7.x, not the Node.js version. If you stick to the default npm that comes with Node.js 14 LTS, you won't have any issues.

ert78gb:
It is nvm for windows https://github.com/coreybutler/nvm-windows.
I always use the default npm version that bundled in the Node. 

> tying down the version so much isn't good IMHO

I develop many packages, application with different node version and `nvm` simplify my life.
You can close this PR if you think it does not help you.

XhmikosR:
It doesn't affect me since I don't use it, but if it helps others, sure. But please use a less restrictive version. Not sure what works here, but something semver-y would be better.

ert78gb:
If we use the exact node.js version then it brings the exact npm version and we can prevent the case that happened in the 
 #447 PR. I installed the dependencies with npm 6.14.11 because it is bundled in the node 14.16.0 and npm added the `"dev": true,` to every dev dependency.

I use nvm every day because I touch 5-10 project every day in my job and it helps to identify which node recommended to development. The `"engine": >=10.12.0`  in the `package.json` so it does not help to identify the recommended node version for dev.

As I mentioned, we can use only the major part of the node version like `14` in the `.nvmrc` file, but different `npm` version bundling in node. The newer version of npm introduces new elements in the `package-lock.json` that not breaking changes just extra info. If developers use different npm version it will generate chaos.

XhmikosR:
Yeah, I'm aware of that so feel free to adapt your PR to use less restrictive versions.

ert78gb:
I close this PR, because if use only the major version of the node in `.nvmrc` file. It does not really help in the current situation.

Issue #446, opened 2021-03-18
v2.0 plan


XhmikosR:
1. replace phantomjs with puppeteer **but make it optional**; we need test cases for this to be safe. We have #284 but needs more work
2. replace image-diff with looks-same; we have #417 but needs some work
3. modernize the codebase; I've made #436, which I try to split it in smaller chunks.
5. (maybe slate this to 2.1.0 or 3.0.0) Run transformations in parallel, see #422. This is more noticeable with thousands of SVG files. Right now svgo runs one at a time.
5. reduce lodash usage to the minimum and then see what we can replace the remaining things with; lodash gets so many vulnerabilities due to its popularity and it's a big package.
6. split tests and streamline the fixture/test images filenames
7. (optional) see what else we can drop from dependencies, for example pretty-diff

yisibl:
At the risk of asking, when are you going to release the first beta version?

XhmikosR:
I think we are now close with phantomjs gone, puppeteer also gone (at least from dependencies). I just want the CI to be stable and I'll cut a new 2.0.0-beta1 release.

The other TODO can probably wait.

yisibl:
This is great, bye PhantomJS and Puppeteer!

XhmikosR:
Any help to fix any other issues is welcome BTW :)

After beta1 and assuming we have managed to remove puppeteer from devDependencies too, I'll update the gulp and grunt packages too.

Kreeg:
@XhmikosR but how to deal with screenshot testing without puppeteer in devDependencies?

XhmikosR:
I'm not sure either. Previously we were using svg2png. It's not maintained a lot, but it worked and it's a lot smaller.

Apart from that, I don't have any ideas, sorry :/

yisibl:
I'm not sure I understand correctly, it should be possible to use [sharp](https://github.com/lovell/sharp) to get a screenshot

And then when resvg supports [`<use>` nesting](https://github.com/RazrFalcon/resvg/issues/6#issuecomment-1038125357) we can use resvg to take screenshots too?


I'm still a little confused, is it accurate to get the size of the SVG through the viewBox and width/height properties? Do you need something like [getBBox()](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox) for the dimensions?

[...]

PR #447, opened 2021-03-20
Replace phantomjs with puppeteer

ert78gb:
closes: #254 
closes #284 by superseding it

Summary:

- puppeteer generates different quality png images than phantomjs
- drop the `svg2png` it is also use phantomjs

TODO:

- [x] Update to the latest puppeteer
- [x] Make puppeteer optional
- [ ] See if we can cache Chromium on CI


XhmikosR:
I don't want to put you off, but... you should really, like really, keep the changes to a minimum. I see tests being split, new ES features used etc, etc. This will be hard to review as it is right now.

ert78gb:
I changed the `"esversion": 6` => `"esversion": 8` in the `.jshintrc` because `6` does not support `async/await`. 
We can use `Promise` instead of `async/await` but in this case I have to write 30% more code and the result will less readable. And 1-2 weeks later we can refactor it to use the newer syntax. #446 `3. modernize the codebase` It is just double work.

> you should really, like really, keep the changes to a minimum

It is the minimal changes to replace `phantomjs`. `phantomjs` for 
- calculate `width` and `height` of the svg if the `viewBow` or `width` and `height` properties missing from the svg file. 
- convert the svg to png.
  -  Once it was implemented in the `capturePhantom` function. 
  - The `svg2png` lib also uses `phantomjs` so I dropped
 
> tests being split, 
Yes because it test the base functionalities  and edge cases of the new code. Put every test in 1 file is not logical for me. This topic is part of the #446 issue `6. split tests and streamline the fixture/test images filenames`. Why should I do double work?

> new ES features used etc, etc
Yes, to write more clear code. The package under a big refactoring why don't use newer ES syntax in the new code base? #446 `3. modernize the codebase`

> This will be hard to review as it is right now.
Which part is hard to review?

XhmikosR:
The problem is that you need to work with me here and try and do this easy for both of us so that we land this and be able to ship v2.0 :)

It's a fact that huge PRs are hard to review. I don't have unlimited time as you probably don't either. So, wait until we land the lint PR. I really need this PR to be as small as possible without any stuff that can be made later, but maybe I was too hard and got carried away by the number of changes. I'll try to review when you are ready and we finish with the other PR.

PS. I agree with async/await instead of Promises of course :)

ert78gb:
I don't have a problem working together with anyone.

After I implemented the major part of this issue I clearly see puppeteer generates different quality png images than phantomjs did so I would like to finish this PR when the #417 landed.

XhmikosR:
#417 isn't ready either, it has its own TODO there. And anyway, image-diff is not the main problem this project has, phantomjs is.

You can always make a PR against the looks-same branch if you want, though.

ert78gb:
@XhmikosR
The color of the zebra changed because the default viewport size of the puppeteer (800px x 600px) and phantomjs is unknown. I did not find what is the default viewport size of the phantomjs, but it should be less than 325px or one of these engine not handle correctly the SVG embended style definitions. Most of the browsers use chromium so maybe this version is good.

XhmikosR:
@ert78gb I updated svgo to the latest version and dropped any custom svgo config. We will need to ignore the zebra changes for the time being and we might as well proceed with #321.

We are getting close with #436, so after that you will need to rebase this branch.

package.json
```diff
@@ -27,11 +27,12 @@
   ],
   "scripts": {
     "lint": "jshint bin lib test",
-    "mocha": "_mocha --reporter spec test/svg-sprite.js",
+    "mocha": "_mocha --reporter spec test/**/*.spec.js test/svg-sprite.js",
```
XhmikosR, line 30: I'd rather be consistent here and use a common filename suffix like .test.js.

package.json
```diff
@@ -27,11 +27,12 @@
...
     "coverage": "c8 npm run mocha",
     "test": "npm run mocha"
   },
   "dependencies": {
+    "@ert78gb/async-init": "^1.0.0",
```
XhmikosR, line 35: Not sure why this is needed. Personally I don't want to add a new dep regardless its size.

package.json
ert78gb: ok, I will rename

package.json
ert78gb: It is responsible for the async singleton creation, the puppeteer browser creation is an async operation. It handles correctly if more than 1 request arriving before the browser created.
If you don't want to add it as a dependency then I have to add files (implementation and test) to this package.

I thought on to create a standalone package for the `BrowserManager` too because it is also reusable.
But in this case, I will not do.

ert78gb: I copied the code from the package to this repo, but not logical for me. I always create small packages if the code useful in other project too. Now we increased the code base and test cases of this package.

package.json
ert78gb: May I rename the `svg-sprite.js` => `svg-sprite.test.js` because of the naming convention?

ert78gb:
@XhmikosR the #417 PR did not properly implement the test file comparison. For [example](https://github.com/svg-sprite/svg-sprite/pull/417/files#diff-6b69305207ae766d26d37a2c2317cd79325033b79362137892c026871dac9116R372)
The old `imageDiff` function always created the diff file, but the `looksSame` does not do it. I think it was a good feature. Don't you wanna add it again?

[OFF]:
[Here](https://github.com/svg-sprite/svg-sprite/pull/417/files#diff-6b69305207ae766d26d37a2c2317cd79325033b79362137892c026871dac9116R148) you run parallel the diff creation and the comparison. It can be a good test speed up function, but 
- if any error happening in the diff creation then the error is absorbed.  
- there was a wish to create the diff only when the images are different. If you don't wait until the result of comparison you don't know should create the diff file or not.
- these callback functions not match with the code modernisation goals

XhmikosR:
> @XhmikosR the #417 PR did not properly implement the test file comparison. For [example](https://github.com/svg-sprite/svg-sprite/pull/417/files#diff-6b69305207ae766d26d37a2c2317cd79325033b79362137892c026871dac9116R372)
> The old `imageDiff` function always created the diff file, but the `looksSame` does not do it. I think it was a good feature. Don't you wanna add it again?

I don't think we need the diff file if everything is fine, there's no need for it.



> [OFF]:
> [Here](https://github.com/svg-sprite/svg-sprite/pull/417/files#diff-6b69305207ae766d26d37a2c2317cd79325033b79362137892c026871dac9116R148) you run parallel the diff creation and the comparison. It can be a good test speed up function, but
> 
>     * if any error happening in the diff creation then the error is absorbed.
> 
>     * there was a wish to create the diff only when the images are different. If you don't wait until the result of comparison you don't know should create the diff file or not.
> 
>     * these callback functions not match with the code modernisation goals

I'm aware of it, hence the TODO in that PR. You are welcome to make a new PR. And please do not mix and match issues.

ert78gb:
> I don't think we need the diff file if everything is fine, there's no need for it.

Yes, it can be useful when the test not pass

> I'm aware of it, hence the TODO in that PR. You are welcome to make a new PR. And please do not mix and match issues.

I think I don't mix. I would like to ask, how do you track the TODO items? 
For example the #417 is merged. It has TODO in the PR, but not created new issue and the source files also not contain any TODO comment.

XhmikosR:
@ert78gb I rebased your branch, could you take care of the 2 remaining TODOs please?

ert78gb:
I will do in the next 2 weeks

ert78gb:
@XhmikosR Can you approve the workflows running. Otherwise I can not check the CI result.
Puppeteer generates a little bit different images on Mac and Linux.

ert78gb:
@XhmikosR I implemented all the 3 TODO.
The Chromium downloading into the `node_modules` folder so it cached.
Download Chromium only when it is used maybe not a good idea because 
- it ~10 sec depend from the network bandwidth and if the user stop the test in the middle of download process the next test run will output a not trivial error message because Puppeteer tries to start a partly downloaded Chromium.
- if the user start 2 svg sprite generation parallel then it will cause a download conflict. Puppeteer does not support cross "application"/script download.

the npm install saves ~10s and 300 MB disk spaces with the optional download.
If the 1 case happening the solution is delete the `node_modules/puppeteer-core/.local-chromium` folder.
if the 2 case the developers have to rewrite their sprite generation process.

I vote on download Chromium with the npm install step because it provide the most straight forward user experience but is it up to you.

package.json
```diff
@@ -28,7 +28,7 @@
   "scripts": {
     "lint": "xo",
     "fix": "npm run lint -- --fix",
-    "mocha": "_mocha --timeout 10000 test",
+    "mocha": "_mocha --timeout 30000 test",
```
XhmikosR, line 31: Can you revert this back to 10 and bump the specific test case to 30s?

XhmikosR:
@ert78gb I thought we could just move puppeteer to the optionalDependencies? That way we could use `try`/`catch` only when the dimensions cannot be determined and the user will need to `npm i puppeteer`.

ert78gb:
please don't do force push because both of us developing on the same branch and it cause unnecessary conflict and mess

package.json
ert78gb: I reverted the setting.

ert78gb:
@XhmikosR I moved puppeteer to the optional dependencies

XhmikosR:
@ert78gb thanks. So, this isn't optimal, I think.

I know there's an [`optionaldependencies`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#optionaldependencies) option and peer dependencies like you used. Not sure which is better in our case, though.

ert78gb:
But puppeteer also needed in production when svg-sprite calculates the dimension of the svg. This is why I thought the peerDependencies are better place

test/browser-mananger.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+        assert.ok(instance1);
+        assert.ok(instance2);
+        assert.deepEqual(instance1, instance2);
+
+        await browserManager.closeBrowser();
+    });
+
+    it('should not throw an exception when the instance does not exist but calls closeBrowser', () => {
+        const browserManager = new BrowserManager();
```
Kreeg, line 22: Some tests are missing `assert` [stage](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) which makes tests less great and may cause problems on CI machine.
Can you please use `assert.doesNotThrow` ? And it will be great to specify expecting error if possible

test/calculate-svg-dimensions.test.js
```diff
@@ -0,0 +1,26 @@
...
+    after(() => browserManager.closeBrowser());
+
+    it('should return the svg dimensions', async() => {
+        const svgFilePath = path.join(__dirname, 'fixture/svg/weather-clear-dimension-calculation.svg');
+        const svg = await readFile(svgFilePath, 'utf-8');
+        const dimensions = await calculateSvgDimensions(svg, await browserManager.getBrowser());
+        const expected = { width: 43, height: 43 };
+
+        assert.deepEqual(dimensions, expected);
+    });
```
Kreeg, line 25: It's all good but where is one moment. Because we treat `calculateSvgDimensions` function like a black box in tests we need to be sure if this function is not returns a hardcoded result of `{ width: 43, height: 43 }`
So, can you please add another fixture with another expected dimensions?
And it'll be great to see a test which checks the consistency of results. So in one test case we expecting that this function returns same result.

test/shape.test.js
```diff
@@ -0,0 +1,46 @@
...
+                assert.equal(Array.isArray(result.shapes), true);
+
+                const svg = result.shapes[0]._contents.toString();
+                const dom = new DOMParser().parseFromString(svg, 'text/xml');
+
+                assert.equal(dom.documentElement.getAttribute('height'), '43');
+                assert.equal(dom.documentElement.getAttribute('width'), '43');
+
+                done();
+            } catch (error) {
```
Kreeg, line 41: This variable is shadowed. Should be changed to something else.

test/svg-sprite.js
```diff
@@ -925,7 +936,7 @@ describe('svg-sprite', () => {
 
                         preview.should.be.ok;
 
-                        capturePhantom(preview, previewImage, error => {
+                        capturePuppeteer(preview, previewImage, error => {
```
Kreeg, line 939: Have to mention that it's cool that you keep same interface here. 👍

Kreeg (changes requested):
@ert78gb hello!
First of all, thank you for such significant amount of work here. I see no critical problems but I myself would suggest some changes in tests.
Do you have time to add some changes I described in review? If not I will do it myself on this weekend

test/browser-mananger.test.js
ert78gb: These tests just validate the concurrency management. You can add assert. the `closeBrowser()` is an async method so I think the `assert.doesNotReject()` is more proper

ert78gb: I don't expect any exception

test/browser-mananger.test.js
```diff
@@ -35,7 +36,7 @@ describe('BrowserManager', () => {
         const browserManager1 = new BrowserManager();
         const browserManager2 = new BrowserManager();
 
-        should.notEqual(await browserManager1.getBrowser(), await browserManager2.getBrowser());
+        assert.notDeepEqual(await browserManager1.getBrowser(), await browserManager2.getBrowser());
```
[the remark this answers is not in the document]
ert78gb: The `notEqual` also good and faster because enough to check the references

test/svg-sprite.js
ert78gb: I tried to minimalise the modifications. I believe in small PRs so if someone wanna improve or rework things they can do it in other PRs.
I think would be good refactor the test because few times hard to figure it out which test failed which pixels are different.

ert78gb:
Hi @Kreeg ,

Thank you for the review. Feel free to finish/extend the PR. If you read through the history maybe you will get the point why I would not like to spend more time on this PR/Project.

test/shape.test.js
```diff
@@ -1,11 +1,13 @@
 'use strict';
 
+const assert = require('assert').strict;
 const { readFileSync } = require('fs');
 const path = require('path');
 const { DOMParser } = require('@xmldom/xmldom');
-const should = require('should');
 const SVGSpriter = require('../lib/svg-sprite.js');
 
+const isObject = obj => typeof obj === 'object' && !Array.isArray(obj) && obj !== null;
```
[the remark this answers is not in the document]
Kreeg: I think it's better to add [lodash.isObject](https://www.npmjs.com/package/lodash.isobject) to project dependencies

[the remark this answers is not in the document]
XhmikosR: I don't think we should keep using lodash anymore and we should in fact try to get rid of it going forward.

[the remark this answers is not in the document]
Kreeg: Well, maybe.
I checked lodash's method and it differs from this. In this case it should stay the same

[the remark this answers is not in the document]
XhmikosR: The current utility works for our case. We should keep it as is for now since it doesn't block this PR.

Later we can try to remove more lodash uses :)

Kreeg:
I removed [weather-clear-dimension-calculation.svg](https://github.com/svg-sprite/svg-sprite/pull/447/commits/91465cddad73e15633d3f0e8a51693d5cc5774d4#diff-9ddd6b0813740b94f91af40618caf4970ae1425dade936f5aabe5639daeabd2d) because it duplicates the `test/svg/special/without-dims.svg` and changed all tests using it.
I also added another one from here https://www.reshot.com/free-svg-icons/weather/. Is it ok? Will it make any problems with copyrights @XhmikosR ? The source does not have any.

Kreeg (approved):
Well this still has some work to do but I'm sure it's non-blocking.

test/browser-mananger.test.js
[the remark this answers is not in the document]
Kreeg: I changed it

Kreeg:
I think we're ready to go here.
@XhmikosR can you please run incomplete checks?

XhmikosR:
I hope puppeteer is made optional later or even see if we can achieve the same with another package because it's an insanely huge package to depend on.

I'll adapt the branch protection rules later.

XhmikosR:
@millette We only need a tool to determine the SVG dimensions if they are not present. Installing puppeteer is ~400MB uncompressed alone on my machine and I don't think it's a good practice with the whole node_modules bloat situation.

Unless we figure out another solution for v2.0.0, I'll have to revert the PR...

Kreeg:
*Sigh* 
I'm checking if resvg can help

Kreeg:
Well, it is kinda working.
The dimensions of `without-dims.svg` is 46x46 (in puppeteer it's 43x43)
The dimensions of `without-dims-2048x2048` is actually 2048x2048
Should I seek deeper? I think it really depends on svg itself rather than on tool.
For example if you open `without-dims.svg` in firefox, the dimensions of glyph itself is 46x46. But in chromium it's 43x43. Which result is right? 
изображение
изображение

XhmikosR:
Do you have another sample? Maybe the one currently used is not the best one?

Kreeg:
I will generate more.

PR #449, opened 2021-03-25
base: no need to trim an empty string

XhmikosR:


jkphl (commented):
@XhmikosR Off the top of my head: I'm not sure if `String()` will take *everything* (especially all sorts of falsy values) and just cast them to a string. If yes, then I'd say we can drop the trailing `|| ''` altogether. I just did that `global || ''` move in order to prevent type errors (but maybe that's not necessary here). I'll try to find some info about what `String()` will be willing to take, but in case you're quicker than me please just go ahead and fix / merge this one.

jkphl:
Ok, I did a few tests and I believe we **cannot** merge this one:

```js
console.log(String(null || '').trim()); // --> ""
console.log(String(null).trim() || ''); // --> "null"
console.log(String(false || '').trim()); // --> ""
console.log(String(false).trim() || ''); // --> "false"
```

So, the current / old behaviour is intentional and we shouldn't change it. I hope it's ok to just close the PR?

XhmikosR:
Yup :)

Still trying to identify places we can improve so I hope you bear with me
for some wrong patches

jkphl:
@XhmikosR Sure — and no worries at all! It's interesting to dig into things like this from time to time, there's always a lot to learn for me. Just keep challenging me! 😉

XhmikosR:
I really missed the fact that the variable can be null etc. Oh, well,
thanks for catching this.

PR #459, opened 2021-04-28
Remove the `_isVinylFile` workaround and use `Vinyl.isVinyl`

XhmikosR:
@jkphl not sure about the JSDoc comments, so please have a look :)


XhmikosR:
@jkphl I don't think these are 100% equivalent. Do you maybe recall any more details?

PR #461, opened 2021-04-28
Replace `_.invoke` with vanilla JS

XhmikosR:
@jkphl this seems to work, but I haven't checked if lodash did exactly the same.

Refs #604

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -265,7 +265,9 @@ SVGSpriter.prototype._compile = function() {
             // Initialize the namespace powers
             while (!this._namespacePow.length || (26 ** this._namespacePow.length < masterShapes.length)) {
                 this._namespacePow.unshift(26 ** this._namespacePow.length);
-                _.invoke(this._shapes, 'resetNamespace');
+                this._shapes.forEach(shape => {
+                    shape.resetNamespace();
+                });
             }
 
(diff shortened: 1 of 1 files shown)


Kreeg:
In original code addressed line of code just do not do anything. Shapes is an array, which has not `resetNamespace` method. So the lodash method here just quietly fails.
I've added test case. You can check yourself that method is not called.
The only way to run it is to call `_.invoke(shapes, '0.resetNamespace')`
Maybe that has worked differently in older versions of lodash.

XhmikosR:
Yeah, I also noticed it was undefined hence why I tried with forEach. Unsure how to proceed though...

Kreeg:
I think we should comment out this whole block and add todo to return to this later.

Kreeg:
@XhmikosR OK, I checked this. I compared the svg's gotten from `npm run example` and `npm run test` commands between `main` and this branches. They are the same.
The real intentions was to call `resetNamespace` on each shape, because shapes are just an array and this method exists only in `SVGShape`. So I assume it is safe. Unit tests here is useless because the previous code did nothing. There is nothing to test except the regression tests, which are OK.

XhmikosR:
@Kreeg thanks! I'm not sure if it's OK TBH, but we could land this and see how it goes.

PR #463, opened 2021-04-28
Replace `_.pick` with native JS

XhmikosR:
Refs #604

### DIFF
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -11,7 +11,6 @@
 
 const fs = require('fs');
 const path = require('path');
-const _ = require('lodash');
 const yaml = require('js-yaml');
 const winston = require('winston');
 const { isFunction, isObject, isPlainObject, isString } = require('./utils/index.js');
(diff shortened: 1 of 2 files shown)


lib/svg-sprite/config.js
```diff
@@ -145,7 +145,8 @@ function SVGSpriterConfig(config) {
             meta = meta ? yaml.load(meta) : {};
             for (const m in meta) {
                 if (_.isPlainObject(meta[m])) {
-                    this.shape.meta[path.join(path.dirname(m), path.basename(m, '.svg'))] = _.pick(meta[m], ['title', 'description']);
+                    const { title, description } = meta[m];
+                    this.shape.meta[path.join(path.dirname(m), path.basename(m, '.svg'))] = { title, description };
```
XhmikosR, line 149: Now that I see this, maybe we could skip the new var and directly access the keys. Let me know what you prefer

Kreeg: I think it's ok here. But I personally prefer skipping such variables. In this case I see no problem

XhmikosR:
@Kreeg if you think any of the PRs are ready feel free to approve. Generally, in the lodash removal PRs, we need to make sure there's no TypeError in case something is undefined/null, but other than that it should be safe.

If you think a PR needs something else please try to push a patch :)

Kreeg:
@XhmikosR Hi!
The problem here is that we can not be 100% sure about these changes since there is no unit tests at all.
IMHO 2.0 release should focus on unit test coverage. Only after it we can do such refactoring.

XhmikosR:
No problem, I have already landed a few changes like this, but I don't mind waiting.

Regardless, this applies to all PRs, some might be safer to land. In these cases, feel free to approve/push further fixes :)

Kreeg (approved):

lib/svg-sprite/config.js
XhmikosR: My only concern is if we end up with undefined `title` and/or `description`. If so, the new code will probably throw.

XhmikosR:
@Kreeg I double checked this and it seems like we are safe, but I could use more eyes.

PR #464, opened 2021-04-28
`readFile`/`readFileSync`: use the string encoding option consistently

XhmikosR:
@jkphl not sure which solution you prefer, both work the same. This is shorter and now consistent.


XhmikosR:
@jkphl friendly ping to have a look at the open PRs.

I'd like to release a v2.0.0-beta1. I'll also try to find some time to cut a new 1.5.1 release but I can't promise about it because we'll go way back in history and CI won't probably work and I lack free time.

PR #501, opened 2021-08-14
Update svgo to v2.4.0

XhmikosR:


XhmikosR:
@TrySound: it seems I messed this up. Could you please have a look and suggest the best way moving forward?

PS. if you need push rights let me know.

example.js
```diff
@@ -11,7 +11,34 @@ const files = glob.sync('**/weather*.svg', { cwd });
...
+    log: 'debug',
+    shape: {
+        transform: [{
+            svgo: {
+                multipass: true,
+                plugins: [{
+                    name: 'preset-default',
+                    params: {
+                        overrides: {
+                            cleanupListOfValues: true,
```
TrySound, line 23: cleanupListOfValues is not part of default preset. It's one of builtin plugins which can be added before or after preset-default plugin. Same with other not default plugins.

XhmikosR: I'm not sure how to proceed. Note that this is an example, the code is in lib/svg-sprite/transform/svgo.js.

I tried to use the master branch on a project and I couldn't get things to work, that's why I asked for help :)

TrySound: What doesn't work?

XhmikosR: This is my test branch: https://github.com/twbs/icons/compare/update-svg-sprite

I notice that `xmlns` are still added. Maybe it's just a stupid thing of wrong merging of the options in this PR...

TrySound: Same removeXMLNS is not part of preset-default

TrySound: I should probably warn about invalid plugins

XhmikosR: Yeah, agreed. A warning would definitely help :)

So, does this mean that one cannot use this plugin anymore? If it's possible, how can one do it? I find the documentation a little unclear.

Thanks for the help!

PS. I hope I'll manage to ship v2.0.0 of svg-sprite which has svgo 2.x, worst case scenario I revert this PR and we'll get the warnings until someone fixes it properly.

XhmikosR: OK, I read again the docs and it's clear how to include it. All good, after all!

I will revisit this PR to make sure all options work.

TrySound:
```
[
  {
    name: 'preset-default',
    params: {
      overrides: {
        removeViewBox: false
      }
    }
  },
  'removeXMLNS'
]
```

XhmikosR: I still can't make `sortAttrs` work with this config:

```json
{
  "mode": {
    "symbol": {
      "dest": ".",
      "sprite": "bootstrap-icons.svg"
    }
  },
  "shape": {
    "transform": [
      {
        "svgo": {
          "multipass": true,
          "plugins": [
            {
              "name": "preset-default",
              "params": {
                "overrides": {
                  "removeAttrs": {
                    "attrs": [
                      "data-name",
                      "fill",
                      "clip-rule"
                    ]
                  },
                  "removeUnknownsAndDefaults": {
                    "keepRoleAttr": true
                  },
                  "removeViewBox": false
                }
              }
            },
            "cleanupListOfValues",
            "removeXMLNS",
            "sortAttrs"
          ]
        }
      }
    ]
  },
  "svg": {
    "namespaceClassnames": false,
    "xmlDeclaration": false
  }
}
```

I should probably move `removeAttrs` outside of `default-plugin` but it seems to work even if it's inside like above.

PR #508, opened 2021-08-21
config: Abstract config code to smaller functions

XhmikosR:
TODO:

- [x] Move `Expand spacing options to arrays`
- [x] Fix JSDoc
- [x] Move `transforms`
- [x] See if we can return early in some places to reduce indentation and make the code clearer


XhmikosR:
@Kreeg After you are done with the config tests (BTW great job as usual!), I think we should look into this PR.

It probably needs more cleanup, so feel free to push here.

Kreeg:
@XhmikosR I think it's ready

PR #535, opened 2021-12-25
config: Fix `xmlDeclaration` and `doctypeDeclaration`

### LINKED ISSUE #523: Fix svgo config after #501
Our options `doctypeDeclaration` and `xmlDeclaration` don't work properly after #501.

korki43:
The options `config.svg.xmlDeclaration` and `config.svg.doctypeDeclaration` will now have these effects:
- `false` -> remove
- `true` -> keep
- empty string -> remove
- non empty string -> keep (will be set to the value later)

This should follow their [documentation](https://github.com/svg-sprite/svg-sprite/blob/9c8a4a6f0830d14e46e2c02542366ed485f88bdc/docs/configuration.md#sprite-svg-options).

It'll use `preset-default` if no svgo plugin configuration is set.

fixes #523
cc @XhmikosR

### DIFF
--- a/lib/svg-sprite/transform/svgo.js
+++ b/lib/svg-sprite/transform/svgo.js
@@ -1,6 +1,5 @@
 'use strict';
 
-
 /**
  * svg-sprite is a Node.js module for creating SVG sprites
  *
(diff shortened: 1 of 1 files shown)


XhmikosR:
Hey, @korki43 sorry for the late reply.

This is what I had in mind too, thanks!

On a side note, it'd be nice if at least the example triggered the warnings but I guess we'll live with it :)

PR #545, opened 2022-02-07
shape.js: use `Object.entries` and `map`

XhmikosR:
This doesn't seem to be covered by tests, so it needs careful review :)

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -341,9 +341,7 @@ SVGShape.prototype._initSVG = function() {
 
     if (entities) {
         let svg = this.svg.current.substr(svgStart.index);
-        for (entity in entityMap) {
-            svg = svg.replace(`&${entity};`, entityMap[entity]);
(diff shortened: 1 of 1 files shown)


lib/svg-sprite/shape.js
```diff
@@ -341,9 +341,7 @@ SVGShape.prototype._initSVG = function() {
 
     if (entities) {
         let svg = this.svg.current.substr(svgStart.index);
-        for (entity in entityMap) {
-            svg = svg.replace(`&${entity};`, entityMap[entity]);
-        }
+        svg = Object.entries(entityMap).map((key, value) => svg.replace(`&${key};`, value));
```
Kreeg, line 344: 1. svg now is an array instead of string
2. `reduce` seems to be more obvious

XhmikosR: You are right of course, didn't check it a lot. I'll skip this for now since there's another place too.

PR #551, opened 2022-02-07
Replace deprecated getter/setter methods with `Object.defineProperty`

XhmikosR:
Not 100% sure about these, needs more eyes :)

### DIFF
--- a/lib/svg-sprite/mode/standalone.js
+++ b/lib/svg-sprite/mode/standalone.js
@@ -89,13 +89,22 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
                 }
             };
 
+            // Create the SVG setter
             Object.defineProperty(this.data.shapes[index], '_svg', {
                 enumerable: false,
                 writable: true
             });
 
-            this.data.shapes[index].__defineSetter__('svg', function(svg) {
-                this._svg = svg;
+            // Create the SVG getter/setter
+            Object.defineProperty(this.data.shapes[index], 'svg', {
+                enumerable: true,
+                configurable: true,
+                get() {
+                    return this._svg;
+                },
+                set(svg) {
+                    this._svg = svg;
+                }
             });
 
             extend(shape, this.data.shapes[index], index);
--- a/lib/svg-sprite/mode/css.js
+++ b/lib/svg-sprite/mode/css.js
@@ -409,16 +409,18 @@ SVGSpriteCss.prototype._addShapeToCSSSprite = function(shape, needsRegular, inde
         writable: true
     });
 
-    this.data.shapes[index].__defineGetter__('svg', function() {
-        return this._svg || shape.getSVG(true, shapeDOM => {
-            for (const r in rootAttributes) {
-                shapeDOM.setAttribute(r, rootAttributes[r]);
-            }
-        });
-    });
-
-    this.data.shapes[index].__defineSetter__('svg', function(svg) {
-        this._svg = svg;
+    // Create the SVG getter/setter
+    Object.defineProperty(this.data.shapes[index], 'svg', {
+        get() {
+            return this._svg || shape.getSVG(true, shapeDOM => {
+                for (const r in rootAttributes) {
+                    shapeDOM.setAttribute(r, rootAttributes[r]);
+                }
+            });
+        },
+        set(svg) {
+            this._svg = svg;
+        }
     });
 };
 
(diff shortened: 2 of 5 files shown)


lib/svg-sprite/mode/standalone.js
```diff
@@ -89,13 +89,20 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
...
             Object.defineProperty(this.data.shapes[index], '_svg', {
                 enumerable: false,
                 writable: true
             });
 
-            this.data.shapes[index].__defineSetter__('svg', function(svg) {
-                this._svg = svg;
+            // Create the SVG getter/setter
+            Object.defineProperty(this.data.shapes[index], 'svg', {
+                get() {
```
XhmikosR, line 100: This might be redundant

Kreeg:
LGTM, but there is more cases in `symbol.js` and `stack.js`

XhmikosR:
Feel free to update the other instances too, IIRC I has some trouble with them (the patch is old, I just rebased it).

lib/svg-sprite/mode/standalone.js
```diff
@@ -89,13 +89,20 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
                 }
             });
 
+            // Create the SVG setter
             Object.defineProperty(this.data.shapes[index], '_svg', {
```
XhmikosR, line 93: Instead we have this here.

lib/svg-sprite/mode/standalone.js
Kreeg: Maybe that was a part of the plan to have only setters here? I'm a little confused

XhmikosR: I was getting an ESLint error (https://eslint.org/docs/rules/accessor-pairs) but happy to disable this one case if we are sure we only need a setter.

Feel free to push to this branch.

Kreeg: It's just more likely that originally it just was forgotten. So in this case it is ok.

Kreeg (approved):

XhmikosR:
I think I'm going to hold off on this for now since we currently don't have tests for all modes.

XhmikosR:
@Kreeg this is now covered but I notice the if conditions are not covered and weren't covered before either. Not sure if something can be improved here.

lib/svg-sprite/mode/standalone.js
```diff
@@ -89,13 +89,22 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
...
                 enumerable: false,
                 writable: true
             });
 
-            this.data.shapes[index].__defineSetter__('svg', function(svg) {
-                this._svg = svg;
+            // Create the SVG getter/setter
+            Object.defineProperty(this.data.shapes[index], 'svg', {
+                enumerable: true,
```
XhmikosR, line 100: Need a double check for these 2 properties

Kreeg: it better to either remove this props or set the default [values](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#description) (false)

XhmikosR: They were needed IIRC.

Kreeg: No, they has default values if not provided. It's not critical here but it adds new functionality here. I'd rather go without it

XhmikosR: Like I said, they are needed:

```
test/svg-sprite/minimal-configuration/modes/defs.test.js (5.687 s)
  ● svg-sprite: default: «defs» mode › creates a visually correct stylesheet resource in CSS format

    TypeError: Cannot redefine property: svg
        at Function.defineProperty (<anonymous>)

      47 |
      48 |         // Create the SVG getter
    > 49 |         Object.defineProperty(dataShape, 'svg', {
         |                ^
      50 |             get() {
      51 |                 return this._svg || shape.getSVG(true, shapeDOM => {
      52 |                     shapeDOM.setAttribute('id', shape.id);
```

Kreeg: Hmm, strange. But OK

XhmikosR: You can try dropping any of the two, maybe I missed something.

XhmikosR: @Kreeg ^^

PR #552, opened 2022-02-07
packer: deduplicate code

XhmikosR:


lib/svg-sprite/mode/css/packer.js
```diff
@@ -46,17 +46,17 @@ SVGSpriteCssPacker.prototype = {};
  * @return {Array} shapes       Packed shapes
  */
 SVGSpriteCssPacker.prototype.fit = function() {
-    const { length } = this.blocks;
-    const width = length ? this.blocks[0].width : 0;
-    const height = length ? this.blocks[0].height : 0;
+    if (this.blocks.length === 0) {
+        return this.positions;
```
XhmikosR, line 50: I opted for returning `this.positions`, unsure if the implicit undefined would be OK.

lib/svg-sprite/mode/css/packer.js
```diff
@@ -46,17 +46,17 @@ SVGSpriteCssPacker.prototype = {};
...
-    this.root.width = width;
-    this.root.height = height;
+    this.root.width = this.blocks[0].width;
+    this.root.height = this.blocks[0].height;
 
-    for (let b = 0; b < length; b++) {
-        const node = this._findNode(this.root, this.blocks[b].width, this.blocks[b].height);
-        const fit = node ? this._splitNode(node, this.blocks[b].width, this.blocks[b].height) : this._growNode(this.blocks[b].width, this.blocks[b].height);
-        this.positions[this.blocks[b].index] = { x: fit.x, y: fit.y };
+    for (const { index, width, height } of this.blocks) {
```
Kreeg, line 56: Maybe switch to
```
this.blocks.forEach(({ index, width, height }) => {...
```

XhmikosR: I prefer `for...of` personally :)

XhmikosR:
I think this patch greatly improves things, especially considering where we started :)

* https://github.com/svg-sprite/svg-sprite/blob/1.5.x/lib/svg-sprite/mode/css/packer.js#L47-L59
* https://github.com/svg-sprite/svg-sprite/blob/main/lib/svg-sprite/mode/css/packer.js#L48-L63
* https://github.com/svg-sprite/svg-sprite/blob/XhmikosR-patch-5/lib/svg-sprite/mode/css/packer.js#L48-L63

lib/svg-sprite/mode/css/packer.js
Kreeg: this should be
```
this.root.width = 0;
this.root.height = 0;
return this.positions;
```

XhmikosR: Not sure if that makes sense? I mean, we are returning `this.positions` so `this.root.*` are not used?

Kreeg: I mean in original code it sets width and height of 0 to this.root

XhmikosR: Yeah, but in the original code we didn't return so we were using `this.root` below. Not sure if it's needed now that we are returning early.

Kreeg: I'm not sure if this will make sense, but since there is no unit testing we can't be sure about this changes :)

Kreeg: @XhmikosR maybe we'll land this PR later after unit testing is done?

XhmikosR: Sure, although I'm not sure I'll have the time to add a test for this.

Kreeg: Should we close it for now?

XhmikosR: Why close it? It's a perfectly fine change that the current tests don't cover. Worst case, we land it as is and see how it goes.

PR #556, opened 2022-02-09
feat: adding tests to cover up svg without dimensions

Kreeg:


test/svg-sprite.test.js
```diff
@@ -215,304 +217,318 @@ describe('svg-sprite', () => {
...
-                        css: {
-                            sprite: 'svg/css.horizontal.svg',
-                            layout: 'horizontal'
-                        }
-                    }, (error, result) => {
-                        result.css.should.be.an.Object;
-                        writeFiles(result).should.be.exactly(1);
-                        svg.horizontal = path.basename(result.css.sprite.path);
-                        done();
+    // Test the minimum configuration
```
Kreeg, line 232: This diff is hard to display but it I simply moved all these tests inside the `forEach` callback function and changed related file names

PR #560, opened 2022-02-11
Replace `_.clone` with Object spread

XhmikosR:
Unsure about the second change (`...element._nsMap`). The first one should be fine since `defaultSVGConfig` is an object.


Kreeg (approved):

XhmikosR:
@Kreeg did you check the second instance so that we are safe? :)

Kreeg:
It'll be an empty object. By the way, there is no `_nsMap` in `Element` declaration. Maybe it was in earliest versions of `xmldom`?

XhmikosR:
Hmm, not sure at all. The codebase wasn't touched until last year I started updating some stuff. We can remove it in a separate PR if you think it's redundant.

Kreeg:
OK

PR #562, opened 2022-02-12
preview: changing comparison of svg images from puppeteer to resvg

Kreeg:
@XhmikosR there is two variants here:
1. Resvg has some problems with displaying srprite without dimensions (this is visible in png changes, the blue svg is corrupted or even dissapeared)
2. There is some problem in our code :D

I'll check the second variant


XhmikosR:
just thinking out loud here... maybe we could even just drop support for SVGs that don't have width/height and/or viewBox and just save us the trouble? Not sure how frequent are such SVGs. Here's how it's done on SVGOMG: https://github.com/jakearchibald/svgomg/blob/3c5f9925f3a8c19f53541ce10299bbc1e5fee7ad/src/js/svgo-worker/index.js#L13-L24

Kreeg:
Then this will be a breaking change. But I'm down with it. I agree with the svgomg's solution. 
SVG without any dimensions is broken and we should just throw an error or warning

XhmikosR:
We can make breaking changes assuming we document it in the release notes.

Not sure the SVGOMG's solution is 100% the same situation, it's probably a little different since SVGOMG doesn't need the dimensions. But we do need them AFAICT.

Kreeg:
@XhmikosR this PR is just a demonstration so I did not run lint and stopped workflow.
I watched these screenshots more closely and noticed changes in normal svg, like `test/expected/png/css.horizontal.mixed.png`. As you can see, two of six svg are missing. This happens with <use> elements inside svg. So it's resvg issue and we should wait until it is fixed.

yisibl:
Thanks for trying resvg-js. For corrupt SVGs, can you submit an issue to the upstream resvg?

Kreeg:
@yisibl i think it's the problem with core package of `rust resvg` package. I'll try.

Kreeg:
@yisibl yeah I'll be never implemented, according to [this](https://github.com/RazrFalcon/resvg/issues/6)

Kreeg:
I think I should try to test screenshot testing via selenium webdriver due to https://github.com/puppeteer/puppeteer/issues/8009
I have never checked the bug-tracker of puppeteer, but when I saw it I was shocked. I think we can't rely on this package :(

PR #564, opened 2022-02-13
fix: fixing valid svg checking

### LINKED ISSUE #454: Valid SVG files are erroneously marked as invalid
An example SVG file that is not able to be processed is [this file](https://github.com/mozilla/pdf.js/blob/6ddc297170a48475b2fdb882efd895d0236474c3/web/images/toolbarButton-pageUp.svg?short_path=aa0160a).
The regex in the following line is not able to match multi-line attribute values of the svg element:
https://github.com/svg-sprite/svg-sprite/blob/7edd0063b0fcf91c436a0d13ec6e01ce01af3299/lib/svg-sprite/shape.js#L324

In the aforementioned example, the `viewBox` attribute of the `svg` element has such a multi-line attribute value:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 
16">
```

Processing this file with the command  `node bin/svg-sprite.js -l debug /path/to/valid/svg` will produce the following log entries:
```
2021-04-05 21:04:48 - debug: Started logging
2021-04-05 21:04:48 - debug: Prepared general options
2021-04-05 21:04:48 - debug: Prepared `shape` options
2021-04-05 21:04:48 - debug: Prepared `svg` options
2021-04-05 21:04:48 - debug: Prepared `mode` options
2021-04-05 21:04:48 - debug: Prepared `variables` options
2021-04-05 21:04:48 - verbose: Initialized spriter configuration
2021-04-05 21:04:48 - debug: Created processing queue instance
2021-04-05 21:04:48 - info: Created spriter instance
2021-04-05 21:04:48 - debug: Added "toolbarButton-pageUp.svg" to processing queue
2021-04-05 21:04:48 - error: Skipping "toolbarButton-pageUp.svg" (Invalid SVG file)
2021-04-05 21:04:48 - info: Compiling 0 shapes ...
2021-04-05 21:04:48 - debug: Created layouter instance
2021-04-05 21:04:48 - info: Finished  sprite compilation
2021-04-05 21:04:48 - info: Created
```

Even though an error occurred while processing the svg file, the error code of the command is 0. The output of the command with flag `-l debug` must be manually inspected to notice that there was an issue.

Kreeg:
+ some JSDoc fixes

Suggestion: moving from `mocha` to `jest`. Current problem is with inability of mocking modules. For example in `test/svg-shape.test.js` I have to copy-paste test input strings from  `test/fix-xml-string.test.js` instead of mocking `lib/svg-sprite/fix-xml-string.js` behavior. 
But there is two problems: 
1. Jest is a BIG package
2. Current test dependency. All test a designed to run step-by-step, but jest is designed to run parallel tests with random orde

Fixes #454

### DIFF
--- a/lib/svg-sprite/fix-xml-string.js
+++ b/lib/svg-sprite/fix-xml-string.js
@@ -1,0 +1,21 @@
+const { DOMParser } = require('@xmldom/xmldom');
+
+module.exports = svgString => {
+    let domParserError = false;
+    const errorHandler = () => {
+        domParserError = true;
+    };
+
+    const fixedSVG = new DOMParser({
+        errorHandler
+    }).parseFromString(svgString).toString().replace(/(\s)(\s+)/g, ' ');
+
+    if (!domParserError) {
+        return fixedSVG;
+    }
+
+    const error = new Error('Invalid XML string');
+    error.name = 'XMLFixingError';
+    error.errno = 1_400_141_400;
+    throw error;
+};
--- a/lib/svg-sprite/fix-xml-string.js
+++ b/lib/svg-sprite/fix-xml-string.js
@@ -6,9 +6,10 @@ module.exports = svgString => {
         domParserError = true;
     };
 
-    const fixedSVG = new DOMParser({
-        errorHandler
-    }).parseFromString(svgString).toString().replace(/(\s)(\s+)/g, ' ');
+    const fixedSVG = new DOMParser({ errorHandler })
+        .parseFromString(svgString)
+        .toString()
+        .replace(/(\s)(\s+)/g, ' ');
 
     if (!domParserError) {
         return fixedSVG;
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -20,6 +20,7 @@ const { CssSelectorParser } = require('css-selector-parser');
 const async = require('async');
 const csso = require('csso');
 const calculateSvgDimensions = require('./calculate-svg-dimensions.js');
+const fixXMLString = require('./fix-xml-string.js');
 
 /**
  * Default callback for shape ID generation
@@ -314,13 +315,29 @@ SVGShape.prototype.setSVG = function(svg) {
  */
 SVGShape.prototype._initSVG = function() {
     // Basic check for basic SVG file structure
-    const svgStart = this.svg.current.match(/<svg(?:\s+[a-z\d-:]+=("|').*?\1)*\s*(?:(\/)|(>[^]*<\/svg))>/i);
+    const validSVGRegExp = /<svg(?:\s+[a-z\d-:]+=(["']).*?\1)*\s*(?:(\/)|(>[^]*<\/svg))>/i;
+    let svgStart = this.svg.current.match(validSVGRegExp);
 
     if (!svgStart) {
-        const e = new Error('Invalid SVG file');
-        e.name = 'ArgumentError';
-        e.errno = 1_429_395_394;
-        throw e;
+        const throwError = () => {
+            const e = new Error('Invalid SVG file');
+            e.name = 'ArgumentError';
+            e.errno = 1_429_395_394;
+            throw e;
+        };
+
+        try {
+            const fixedXMLString = fixXMLString(this.svg.current);
+            svgStart = fixedXMLString.match(validSVGRegExp);
+
+            if (!svgStart) {
+                return throwError();
+            }
+
+            this.svg.current = fixedXMLString;
+        } catch {
+            throwError();
+        }
     }
 
     // Resolve XML entities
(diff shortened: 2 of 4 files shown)


test/fix-xml-string.test.js
```diff
@@ -0,0 +1,39 @@
+const assert = require('assert').strict;
+const fixXMLString = require('../lib/svg-sprite/fix-xml-string.js');
+
+describe('testing fixing svg string', () => {
+    it('should return valid svg file on svg with multiline attribute values', () => {
+        assert.equal(fixXMLString(`<svg viewBox="0 0 0 16
```
XhmikosR, line 6: Are you sure these are valid SVGs? SVGOMG doesn't parse it either, so SVGO will not parse it either.

Kreeg: Yep, I checked it out in online XML and SVG validators

XhmikosR:
OK I start to believe there's indeed something wrong with the tests. They seem to fail frequently in `calculateSvgDimensions`. Something weird is going on there or the Windows runners are slow on many occasions.

We should try to increase mocha's timeout to see if it fixes the issues.

Unrelated to this PR, though, we just need a steady CI :)

Kreeg:
@XhmikosR I'll check. I think it's regardless of `calculateSvgDimensions` test and not depends on tests order. We'll see

Kreeg:
@XhmikosR I've changed the order of the tests and it working now in 100% times.
I am sure this isn't appropriate solution, but it make CI stable again. I am not sure what was causing the problem, but it surely not `calculateSvgDimensions` itself

XhmikosR:
@Kreeg can you make a separate PR for the tests fix please? And maybe choose better names, maybe numbers if a specific order is needed? Although, I don't quite get why this is needed.

Kreeg:
@XhmikosR I did it just to demonstrate that is `calculateSvgDimension` is not the problem itself but the real reason is somewhere around.
I'll dig it. For now I'll rollback my changes.
What I mean here is what the changing the order of tests should not be a solution. This makes it weak and hardly and non-obviously intertwined.

Kreeg:
@yisibl what do you think? This problem appeared after the resvg was added. The tests start failing by timeout (10s) randomly on windows machines with node v12, v14, v16

test/fix-xml-string.test.js
yisibl: An SVG file must have a `xmlns="http://www.w3.org/2000/svg"` namespace. See https://github.com/RazrFalcon/resvg/issues/192#issuecomment-569467534

If namespace is missing, it will report an error in resvg. I added a [test case in resvg-js](https://github.com/yisibl/resvg-js/pull/64).

yisibl:
> @yisibl what do you think? This problem appeared after the resvg was added. The tests start failing by timeout (10s) randomly on windows machines with node v12, v14, v16

For test timeout, try increasing the test timeout.

XhmikosR:
I the issue is solved by changing the order of tests, it means it's not a timeout issue but rather an issue with a test actually failing and as a result hitting the timeout without success.

XhmikosR:
All right, the Windows issue will be fixed by #567. It's unrelated to resvg or anything.

Kreeg:
@yisibl thanks for the answer! Sorry to bother :)

test/fix-xml-string.test.js
Kreeg: @yisibl hey! Thank!
This test is just for `fix-xml-string.js` which just fixes styles and incompletion in provided strings.

lib/svg-sprite/fix-xml-string.js
```diff
@@ -0,0 +1,22 @@
...
+        .parseFromString(svgString)
+        .toString()
+        .replace(/(\s)(\s+)/g, ' ');
+
+    if (!domParserError) {
+        return fixedSVG;
+    }
+
+    const error = new Error('Invalid XML string');
+    error.name = 'XMLFixingError';
```
XhmikosR, line 19: Does it still make sense to keep using custom Error Objects? I mean, we don't test them and it's probably enough if we have a clear error message only anyway.

test/fix-xml-string.test.js
yisibl: OK

lib/svg-sprite/shape.js
```diff
@@ -314,13 +315,29 @@ SVGShape.prototype.setSVG = function(svg) {
  */
 SVGShape.prototype._initSVG = function() {
     // Basic check for basic SVG file structure
-    const svgStart = this.svg.current.match(/<svg(?:\s+[a-z\d-:]+=("|').*?\1)*\s*(?:(\/)|(>[^]*<\/svg))>/i);
+    const validSVGRegExp = /<svg(?:\s+[a-z\d-:]+=(["']).*?\1)*\s*(?:(\/)|(>[^]*<\/svg))>/i;
```
XhmikosR, line 318: So, the only change in the regex is `"|'` to `["']`?

XhmikosR: I thought we would need to make the regex a multiline one with the `m` flag.

Kreeg: @XhmikosR yes, the only change is that.
Just changing the flag of this regexp did not help, this tests are failing

XhmikosR (approved):

PR #572, opened 2022-02-23
feat: moving from mocha to jest

### LINKED ISSUE #445: Make each test case individually testable
Yesterday I tried to investigate the #421 , but very hard to debug the tests because the `it.only` not working. I have to find wich describe section contains the prerequisites. I think in long term worth refactoring the test to use `beforeEach` to setup the test. It will increase the running time, but it will allow running parallel so overall we can decrease the test time.

IMHO I think worth splitting the test into multiple files. Maybe the good separation is the `mode` but currently I don't have enough code knowledge to identify it.

Kreeg:
Pros:

- complete package
- integrated parallel multi-worker tests runner
- integrated code coverage (and more accurate than c8, can be confirmed by running `npm run coverage:jest` and `npm run coverage:jest:native`). Bye c8.
- integrated stubs, mocks (bye sinon)
- integrated assertion library (bye `should`, `assert`)
- great and big community with a lot of plugins
- es modules ready (code transform with babel-jest)
- integrated watcher `jest --watch`

Cons:
- this is a BIG package
- the first run may took a long time (if the code transformation is included)
- may need an adjustments with current github actions

TODO:

- [x] remove unused packages and files
- [x] configure Jest's coverage to include all files in bin and lib folder, and use the same reporters as before
- [x] see if you can split the async compile in a separate PR since it's a feature that doesn't really belong to the Jest changes
- [x] see if the root folder is good practice to keep the jest files
- [x] remove the constants file since it's pretty much useless and we don't need to test that string IMHO
- [x] fix exceeding timeout on failed tests

Fixes #445 along the way


XhmikosR:
I personally have no interest in this so I'm +0. But if it helps you move easier, I don't mind the switch.

I could also argue about some of the pros you list, but none of us will benefit from this ;)

Bottom line, I don't mind moving to Jest, but this PR is far from ready AFAICT.

Kreeg:
Yeah I just wanted to draw your attention to this and to get your opinion. Thanks for reply I'll move forward with this.

lib/svg-sprite/constants.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    DEFAULT_XML_DECLARATION: '<?xml version="1.0" encoding="utf-8"?>'
```
XhmikosR, line 2: This seems useless just for one case IMHO.

Kreeg:
JEST
```
Test Suites: 12 passed, 12 total
Tests:       104 passed, 104 total
Snapshots:   0 total
Time:        27.024 s, estimated 29 s
```
MOCHA
```
 113 passing (47s)
```

Jest Coverage (native)
```
All files                     |   85.07 |    64.39 |   88.05 |   85.25 |
```

Jest coverage (c8)
```
All files                     |   81.39 |    77.11 |    90.9 |   81.39 |
```

Mocha Coverage (с8)
```
All files                     |    81.1 |    76.16 |   88.69 |    81.1 |
```

Where is view.mixed.test.js, view.packed.test.js left.
NOTE: shape.test.jest.js was changed because of mocking the results, and thus getting rid of obsolete tests

package.json
```diff
@@ -29,8 +29,12 @@
     "fix": "xo --fix",
     "mocha": "mocha \"test/**/*.test.js\"",
     "coverage": "c8 npm run mocha",
+    "coverage:jest": "c8 npm run test:jest",
```
XhmikosR, line 32: let's just keep jest's default coverage and remove c8 dev dependency

XhmikosR:
@Kreeg when you are ready let me know so that I can have a closer look :)

XhmikosR:
BTW one things that bothers me on main is that after running each test file, the tmp folder is cleaned. So, if one test suite fails, I have to explicitly run the failed one again so that I can get the updated PNG files from tmp.

I think we should clear the tmp folder via cli or just don't clear it at all.

package.json
Kreeg: Ok but it is still in draft :D

Kreeg:
@XhmikosR Ok I'll tell you when it is ready
But I am not quite follow you on deleting paths. `tmp` path is removed before all test are ran, each test has his own directory inside `tmp` and is cleaned before the tests inside the current test suite.
There is no teardown in which tmp path is deleted so it is exists after tests are passed up till the next run

XhmikosR:
No worries, we'll check it out again later :)

Kreeg:
@XhmikosR Hey! The first step is done. I kept old test files for visual comparing with the new ones and to have ability to run old tests. 
The next step will be removing useless dependencies and updating scripts in `package.json`

jest.setup.js
```diff
@@ -0,0 +1,110 @@
...
+
+                resolve({ isEqual: result.equal, difference: JSON.stringify(result.diffClusters) });
+            });
+        });
+    });
+};
+
+// eslint-disable-next-line jest/require-hook
+expect.extend({
+    async toBeVisuallyEqual(receivedSVGPath, expectedPNGPath) {
```
Kreeg, line 51: Maybe this functions should be renamed, please suggest if so.

lib/svg-sprite/constants.js
Kreeg: Maybe, but at the same time it adds more lucidity where it used.

package.json
Kreeg: I think this should be done in another PR right after this one is merged. Otherwise this PR will be HUGE. But if it is ok I will change it after tests review

test/calculate-svg-dimensions.test.jest.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+
+const { readFileSync } = require('fs');
+const path = require('path');
+const calculateSvgDimensions = require('../lib/svg-sprite/utils/calculate-svg-dimensions.js');
+
+describe('calculateSvgDimensions', () => {
+    it('should return the expected dimensions from 46x46 fixture', () => {
+        expect.hasAssertions();
```
Kreeg, line 9: This is for the eslint-plugin-jest. The point is if the test has no assertions at all it will be marked as passed no matter what it does nothing. It added false confidence in tests. 
This eslint rule guards for such errors

test/utils.test.jest.js
```diff
@@ -0,0 +1,199 @@
+'use strict';
+
+/* eslint-disable unicorn/new-for-builtins, no-new-wrappers, prefer-regex-literals, jest/prefer-expect-assertions */
```
Kreeg, line 3: In the same time some test cases will look as 
```
expect.hasAssertions();
expect....
```
In this case we can ignore that rule. Or copy-paste `expect.hasAssertions();` in all tests. All the variants are good

lib/svg-sprite/constants.js
XhmikosR: I'd still drop it. Unless, the file has a clear reason of existence, better drop it.

lib/svg-sprite.js
```diff
@@ -220,6 +220,23 @@ SVGSpriter.prototype.compile = function(config, cb) {
     this._compile();
 };
 
+/**
+ *
+ * @param {object} config                                   Configuration
+ * @returns {Promise<{result:{object}, data:{object}}>}     Results
+ */
+SVGSpriter.prototype.compileAsync = function(config) {
```
XhmikosR, line 228: This is new. Should be added in a new branch, or at least a clear patch, but I'd still prefer a separate PR so that it's clearly stated in the release notes.

package.json
XhmikosR: Well, as this PR is right now, it's incomplete. It's not being used on CI and it's not final. I'd say better make all the changes related to the switch to jest here and remove everything that's not needed, but keep the old npm scripts.

test/utils.test.jest.js
XhmikosR: Not sure I follow what you mean?

jest.setup.js
XhmikosR: this one is OK. But can we move the jest files in test folder? Not sure what's the common practice here.

XhmikosR:
I added these in the OP:

- [ ] remove unused packages and files
- [ ] configure Jest's coverage to include all files in bin and lib folder, and use the same reporters as before

XhmikosR:
Great job BTW! Even though I don't use Jest a lot, I see its benefits. We might also want to disable verbose output.

XhmikosR:
Actually, it's OK if you think we should remove the old packages and stuff in a separate PR.

But, we should still try to fix these:

1. configure Jest's coverage to include all files in bin and lib folder, and use the same reporters as before
2. disable verbose output
3. see if the root folder is good practice to keep the jest files
4. remove the constants file since it's pretty much useless and we don't need to test that string IMHO
5. see if you can split the async compile in a separate PR since it's a feature

package.json
Kreeg: OK

lib/svg-sprite/constants.js
Kreeg: Ok

jest.setup.js
Kreeg: @XhmikosR it can be placed inside another folder, yes. The config file may too, but in this case we need to pass it's location in script like `jest --config=path`

test/utils.test.jest.js
Kreeg: @XhmikosR I am talking about this rule https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-expect-assertions.md. I commented about it here https://github.com/svg-sprite/svg-sprite/pull/572#discussion_r825446342
So there is two options
1. `expect.hasAssertions` in every test case
2. disabling this rule in some test cases like in this one

lib/svg-sprite.js
Kreeg: Ok! I'll do it tomorrow

Kreeg:
@XhmikosR thanks! I'm actually reconsider removing old tests and dependencies in another PR. This PR should be the complete shifting (not the one step)

jest.setup.js
XhmikosR: Would be OK if we kept the config in root and moved the other files maybe?

test/utils.test.jest.js
XhmikosR: Hmm, both work for me but let's try to keep it simple for future contributors :)

jest.config.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+module.exports = {
+    clearMocks: true,
+    resetMocks: true,
+    coverageProvider: 'v8',
+    coverageReporters: ['html', 'lcov', 'text'],
```
XhmikosR, line 7: This seems to create two folders in coverage, unsure if we can change something.

test/svg-sprite/minimal-configuration/modes/css.test.js
```diff
@@ -97,25 +97,19 @@ describe('testing minimal config', () => {
             // Horizontal layout
             it('horizontal layout', async() => {
                 expect.hasAssertions();
-                await expect(path.join(tmpPath, 'css/svg', svg.horizontal)).toBeVisuallyEqual(
-                    path.join(paths.expectations, `png/css.horizontal${testConfig.namespace}.png`)
-                );
+                await expect(path.join(tmpPath, 'css/svg', svg.horizontal)).toBeVisuallyEqual(path.join(paths.expectations, `png/css.horizontal${testConfig.namespace}.png`));
```
XhmikosR, line 100: @Kreeg let's split such cases in 2 variables, like `input`, `expected` or something like that in all cases. It should make things more clear and we can get rid of the long lines :)

test/helpers/compare-svg-2-png-helper.js
```diff
@@ -16,14 +16,14 @@ const convertSvg2Png = require('./convert-svg-2-png.js');
  * @param {Function} done             Callback
  */
 module.exports = async(svg, png, expected, diff, done) => {
-    fs.mkdirSync(path.dirname(png), { recursive: true });
+    await fs.mkdir(path.dirname(png), { recursive: true });
```
XhmikosR, line 19: There are more cases we could/should use async but we can tackle them later :)

test/shape.test.js
```diff
@@ -23,45 +23,48 @@ const expectations = [{
...
-                    done(compilationError);
-                }
-            });
-        });
+        expect(result).toBeInstanceOf(Object);
+        expect(result.shapes).toBeInstanceOf(Array);
+
+        const dom = new DOMParser().parseFromString(result.shapes[0]._contents.toString(), 'text/xml');
+
+        expect(dom.documentElement.getAttribute('height')).toStrictEqual(dimension.height.toString());
```
XhmikosR, line 67: is there any way to have jest always use strict assertions like we had `assert.strict`?

Kreeg: This is a strange case here, here is the rules https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-strict-equal.md, https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be.md.
`toStrictEqual` is used to compare objects (with structure), otherwise it is preferred to use `toBe`.
In this case I think its linting problem (because it can't get type of `getAttribute`) and we can change it to `toBe`, that's better

Kreeg: Anyway, preferring strict mode is already covered by eslint

XhmikosR: I wasn't talking about the `'use strict' directive 😛

I was wondering why you were using `toStrictEqual` since `toEqual` exists too and if we can always enable strict checks like assert has with strict mode. :)

XhmikosR:
@Kreeg I think these only are left:

1. see if there are more instances of `toStrictEqual` and similar with `toBe` or `toEqual` when possible (if the remaining instances need to be strictEqual, we are fine :))
2. maybe split those long lines to expected/input variables (I'll have another look myself later in case we missed something)

They are not blocking of course, it's just nice to have before we land this huge PR :)

I also have a question about the coverage but I'm not sure if it's a misconfiguration issue. I see duplicate files in `coverage/` and in `coverage/lcov-report/` folders.

Again, great work!

XhmikosR (approved):

test/calculate-svg-dimensions.test.js
```diff
@@ -1,66 +1,76 @@
 'use strict';
 
-const assert = require('assert').strict;
-const { readFileSync } = require('fs');
+const { readFile } = require('fs').promises;
 const path = require('path');
 const calculateSvgDimensions = require('../lib/svg-sprite/utils/calculate-svg-dimensions.js');
 
-const { paths } = require('./helpers/constants.js');
```
XhmikosR, line 8: Makes me wonder if we could/should just remove helpers/constants.js later?

Kreeg: We'll see. Maybe we can reduce them, but removing all of them seems impossible for now.

Kreeg:
@XhmikosR 
1. There is no `toEqual` now, only `toStrictEqual` for the objects and `toBe` for the rest.
2. Can you please specify this lines? The problem is only with the length? This can be easily auto-fixed with prettier ( I know, it's me again with that prettier, but we can turn it on in tests folder for now). If length is not the case (more readability) we can refactor to use `expect(input).to...(expected)` in this PR.
3. I faced now issue with the failed tests. They are all fall by timeout. I'll check it. That's critical

XhmikosR:
Re 1., 2. all good now :)

About 3., it's not new it's the same issue we had in the past. 15s might work better.

Kreeg:
@XhmikosR yeah `createDiff` takes too long. So I increased timeout.

Kreeg:
About coverage folders. The files are not quite the same. I think it's jest "feature". Is it a problem?

XhmikosR:
Great, let's land it and we can tweak it later!

1. more async in tests (and later in core)
2. see if we can figure out the duplicate coverage issue

Kreeg:
I'm not sure which is correct too. I think they are all correct, but looks different (some files are not present in lcov directory)

XhmikosR:
Can you make a branch adding your preferred prettier options but trying to match the current style?

Assuming we have a prettier config file, we should be able to make use of it in the whole codebase and for JS we can use xo + prettier.

Kreeg:
@XhmikosR OK! But I think to match all this styles with prettier is impossible but I'll try

XhmikosR:
Yeah, we'll need to adapt some things for sure but it should be doable :)

PR #573, opened 2022-02-26
fix: changing async to sync render with resvg

Kreeg:
That is for increasing speed on win32 machines in CI


Kreeg:
I added timeout for the first test for now

PR #574, opened 2022-02-26
preheat resvg rendering in tests

Kreeg:


Kreeg:
@yisibl hello!
What do you think about this? The first render of resvg on win32 machine took more than 15 secs sometimes. This forced me to add preheating resvg with rendering dummy svg outside of the test blocks.
It happened about 95% of time.

yisibl:
Is this file very slow? https://github.com/svg-sprite/svg-sprite/blob/main/test/fixture/svg/special/without-dims/46x46.svg

test/helpers/resvg-preheat.js
```diff
@@ -0,0 +1,7 @@
+// resvg preheat
+const resvg = require('@resvg/resvg-js');
+
+resvg.render('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"></svg>', {
+    fitTo: { mode: 'original' },
```
yisibl, line 5: This is the default value, please remove this configuration.

Kreeg: Ok thanks

Kreeg:
@yisibl no it is not the file itself. The actual rendering time is now about 46ms, the problem was with the **first** rendering of any file on win32 machines.

yisibl:
Please try locking resvg-js version to 1.0.4, and let's see if the problem is introduced by a new version of resvg using the exclusion method.

PR #575, opened 2022-02-27
feat: test refactoring

Kreeg:


test/helpers/add-files.js
```diff
@@ -0,0 +1,47 @@
+const fs = require('fs');
```
XhmikosR, line 1: Please add `'use strict'` statements + a new line in all new files (this is not enforced in lint because the files are treated as modules)

PR #576, opened 2022-02-28
feat: disable loadSystemFonts in resvg-js to improve first-time loading speed

yisibl:
Try to fix: https://github.com/svg-sprite/svg-sprite/pull/574#issuecomment-1053551158

resvg-js loads all system fonts by default, which is more time consuming on first startup and is cached after startup, which explains why preloading resvg-js tests once makes them faster.

We don't need this feature by default, so disable this option. This can greatly improve the speed of svg-sprite the first time it is started.

This PR also removes `fitTo: { mode: 'original' }` option, as it is the default value.

### DIFF
--- a/lib/svg-sprite/calculate-svg-dimensions.js
+++ b/lib/svg-sprite/calculate-svg-dimensions.js
@@ -11,8 +11,10 @@ const imageSize = require('image-size');
 function calculateSvgDimensions(svg) {
     try {
         const pngData = resvg.render(svg, {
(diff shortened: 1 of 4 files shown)


yisibl:
@Kreeg, Please take a look, I can't continue to run CI.

Kreeg:
@yisibl wow thank you! I'll check tonight

Kreeg (approved):

Kreeg:
@yisibl thanks again!

PR #577, opened 2022-03-01
Substitute ID references in href attributes

p-a:
Only xlink:xhrefs were substituted, SVG 2 also allows for href

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/href

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -711,6 +711,14 @@ SVGShape.prototype.setNamespace = function(ns) {
                 }
             });
 
+            // Substitute ID references in href attributes
+            select('//@href', this.dom).forEach(href => {
+                const hrefValue = href.nodeValue;
+                if (!hrefValue.startsWith('data:') && hrefValue in substIds) {
+                    href.ownerElement.setAttribute(href.name, `#${substIds[hrefValue]}`);
+                }
+            });
+
             // Substitute ID references in referencing attributes
             svgReferenceProperties.forEach(refProperty => {
                 select(`//@${refProperty}`, this.dom).forEach(ref => {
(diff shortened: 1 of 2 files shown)


lib/svg-sprite/shape.js
```diff
@@ -708,6 +708,14 @@ SVGShape.prototype.setNamespace = function(ns) {
                 }
             });
 
+            // Substitute ID references in href attributes
```
Kreeg, line 711: Please change this comment to what it really does

Kreeg:
Hello! Please provide some tests to cover addressed changes

lib/svg-sprite/shape.js
p-a: Well, it is basically the same thing as the function above which is only for xlink:href. Which has the comment
`                // Substitute ID references in xlink:href attributes`

p-a:
Added some tests. Didn't find any for the xlink:href substitution so I added one for that too

test/svg-sprite/shape/shape.svg.test.js
```diff
@@ -119,4 +119,55 @@ describe('testing getSVG()', () => {
...
+                },
+                mode: {
+                    view: true,
+                }
+            },
+            verbose: jest.fn()
+        });
+        shape.complement(() => {});
+        shape.setNamespace('ns-');
+        expect(shape.getSVG()).toBe('<svg id=\"someprefix-ns-abc\" height=\"0\" width=\"0\" viewBox=\"0 0 0 0\"><use xlink:href=\"#someprefix-ns-abc\"/></svg>');
```
Kreeg, line 170: there is unnecessary backslashes here, before the quote marks

lib/svg-sprite/shape.js
```diff
@@ -711,6 +711,14 @@ SVGShape.prototype.setNamespace = function(ns) {
...
+            select('//@href', this.dom).forEach(href => {
```
Kreeg, line 715: Maybe this selector should be more specific to select only needed elements? Like [here](https://stackoverflow.com/a/46274695/1420159)

Kreeg:
@p-a Hello! Have you ran `npm run fix`?

Kreeg:
Closed due to inactivity and impossibility to push to source branch

PR #583, opened 2022-03-02
feat: test refactoring

Kreeg:
My point here to split tests by logical peaces.
One tests creation of files 
Second one just write those files before the actual visual regression tests

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,5 +1,5 @@
 {
-  "timeout": 15000,
+  "timeout": 10000,
   "check-leaks": true,
   "throw-deprecation": true,
   "trace-deprecation": true,
--- a/test/svg-sprite/minimal-configuration/rerun.test.js
+++ b/test/svg-sprite/minimal-configuration/rerun.test.js
@@ -1,0 +1,71 @@
+'use strict';
+
+/* eslint-disable no-unused-expressions, max-nested-callbacks */
+const path = require('path');
+const glob = require('glob');
+const should = require('should');
+const SVGSpriter = require('../../../lib/svg-sprite.js');
+const { addFixtureFiles } = require('../../helpers/add-files.js');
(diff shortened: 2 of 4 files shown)


.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "timeout": 15000,
+  "timeout": 10000,
```
Kreeg, line 2: I also turned it down because now resvg is doing fine

test/svg-sprite/minimal-configuration/rerun.test.js
```diff
@@ -0,0 +1,67 @@
...
+                layout: 'vertical',
+                dimensions: true,
+                render: {
+                    css: true,
+                    scss: true,
+                    less: true,
+                    styl: true
+                }
+            }
+        }, async(_, firstResult) => {
```
XhmikosR, line 34: Why ignore error?

XhmikosR:
Looking good, assuming everything is still checked. 👍 

The less IO, the better, although if we use async methods it shouldn't matter so much :)

test/svg-sprite/minimal-configuration/rerun.test.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+/* eslint-disable no-unused-expressions, max-nested-callbacks */
+const path = require('path');
+const glob = require('glob');
+const should = require('should');
```
XhmikosR, line 6: BTW I planned at some point to get rid of should and just use assert, but never got to do it :)

Kreeg: Yeah it will be eliminated in jest version of tests :)

test/svg-sprite/minimal-configuration/rerun.test.js
Kreeg: An error, thanks

Issue #570, opened 2022-02-22
Increased load time on main 


XhmikosR:
1.5.4:

```
C:\Users\xmr\Desktop\svg-sprite>git clean -dfx && npm i --production
Removing node_modules/
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated phantomjs-prebuilt@2.1.16: this package is now deprecated
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated svgo@1.3.2: This SVGO version is no longer supported. Upgrade to v2.x.x.

added 241 packages, and audited 242 packages in 5s

38 packages are looking for funding
  run `npm fund` for details

3 moderate severity vulnerabilities

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.

C:\Users\xmr\Desktop\svg-sprite>hyperfine "node ../load-time.js"
Benchmark 1: node load-time.js
  Time (mean ± σ):     381.6 ms ±  24.6 ms    [User: 1.3 ms, System: 5.9 ms]
  Range (min … max):   361.2 ms … 439.9 ms    10 runs
```

main:

```
C:\Users\xmr\Desktop\svg-sprite>git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.

C:\Users\xmr\Desktop\svg-sprite>git clean -dfx && npm i --production
Removing node_modules/

added 102 packages, and audited 103 packages in 3s

12 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

C:\Users\xmr\Desktop\svg-sprite>hyperfine "node ../load-time.js"
Benchmark 1: node load-time.js
  Time (mean ± σ):     531.8 ms ±  23.6 ms    [User: 1.4 ms, System: 5.0 ms]
  Range (min … max):   508.1 ms … 577.0 ms    10 runs
```

load-time.js:

```js
'use strict';

console.log('Load time');
console.time('Module');

exports.m = require('./svg-sprite');

console.timeEnd('Module');

```

I haven't been able to pinpoint the root cause yet, but it's a significant increase we should try to figure out.

XhmikosR:
[...]

PR #585, opened 2022-03-03
Move csso require to improve load time

XhmikosR:
@Kreeg this saves approximately ~100ms on my machine. Unsure if there's a better way, but this seems to do the job.

Refs #570


XhmikosR:
Although, the whole csso block is untested, I don't even know if it works and/or if we could require only the parts we need to.

Kreeg (approved):
LGTM!

PR #590, opened 2022-03-03
feat: decrease Lodash usage

peterblazejewicz:
As per comment in v2 beta release this decreases direct usage of Lodash:

> Reduced lodash usage (patches welcome to reduce its use further!)

- isFunction
- isObject

Additional gain should be a slight speed increase.

Thanks!

TODO:

- [x] tighten `isObject`
- [x] add tests for `isPlainObject`

### DIFF
--- a/lib/svg-sprite/utils/index.js
+++ b/lib/svg-sprite/utils/index.js
@@ -5,16 +5,12 @@
  * @returns {boolean}    Returns true if value is correctly classified, else false.
  */
 function isFunction(value) {
-    if (value && typeof value === 'function') {
-        return true;
-    }
-
-    return false;
+    return Boolean(value && typeof value === 'function');
 }
 
 /**
- * Checks if value is the language type of Object. (e.g. arrays, functions, objects, regexes, new Number(0),
- * and new String(''))
+ * Checks if value is the language type of Object (e.g. objects, regexes, new Number(0),
+ * and new String('')). Excluding arrays (new Array())
  *
  * @param {any} value The value to check.
  * @returns {boolean} Returns true if value is an object, else false.
@@ -27,7 +23,7 @@ function isObject(value) {
  * Checks if value is an Object
  *
  * @param {any} value The value to check.
- * @returns {boolean} Returns true if value is an object, else false.
+ * @returns {boolean} Returns true if value is an plain object, else false.
  */
 function isPlainObject(value) {
     return Object.prototype.toString.call(value) === '[object Object]';
--- a/lib/svg-sprite/utils/index.js
+++ b/lib/svg-sprite/utils/index.js
@@ -33,8 +33,19 @@ function isPlainObject(value) {
     return Object.prototype.toString.call(value) === '[object Object]';
 }
 
+/**
+ * Checks if value is a String
+ *
+ * @param {any} value The value to check.
+ * @returns {boolean} Returns true if value is a String, else false.
+ */
+function isString(value) {
+    return Object.prototype.toString.call(value) === '[object String]';
+}
+
 module.exports = {
     isFunction,
     isObject,
-    isPlainObject
+    isPlainObject,
+    isString
 };
--- a/lib/svg-sprite/utils/index.js
+++ b/lib/svg-sprite/utils/index.js
@@ -20,7 +20,7 @@ function isFunction(value) {
  * @returns {boolean} Returns true if value is an object, else false.
  */
 function isObject(value) {
-    return value instanceof Object;
+    return typeof value === 'object' && !Array.isArray(value) && value !== null;
 }
 
 /**
--- a/lib/svg-sprite/utils/index.js
+++ b/lib/svg-sprite/utils/index.js
@@ -1,11 +1,11 @@
 /**
  * Checks if value is a callable function.
  *
- * @param {string} f The value to check.
+ * @param {string} value The value to check.
  * @returns {boolean}    Returns true if value is correctly classified, else false.
  */
-function isFunction(f) {
-    if (f && typeof f === 'function') {
+function isFunction(value) {
+    if (value && typeof value === 'function') {
         return true;
     }
 
--- a/lib/svg-sprite/utils/index.js
+++ b/lib/svg-sprite/utils/index.js
@@ -1,0 +1,29 @@
+/**
+ * Checks if value is a callable function.
+ *
+ * @param {string} f The value to check.
+ * @returns {boolean}    Returns true if value is correctly classified, else false.
+ */
+function isFunction(f) {
+    if (f && typeof f === 'function') {
+        return true;
+    }
+
+    return false;
+}
+
+/**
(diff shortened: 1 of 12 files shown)


XhmikosR:
Thanks for the PR!

I have some comments about the implementation.

1. I'd just use one utils file
2. I prefer not to use regex, even if it's more verbose

```js
function isFunction(f) {
    if (f && typeof f === 'function') {
        return true;
    }

    return false;
}

function isString(string) {
    if (string && typeof string.valueOf() === 'string') {
        return true;
    }

    return false;
}
```

peterblazejewicz:
no worries, I'll amend. I've used multifile setup, in case it could be useful for ESM. to be amended

XhmikosR:
I'd switch to ESM too, but I had issues with the grunt/gulp packages. So, I'd say ESM is something for a later time (unknown :P).

peterblazejewicz:
amended

> BTW there are more instances of at least isObject in bin.

using VSCode, when creating default 'exclude' for IDEs features, the 'bin' is there by default IMO, this is why I've missed this.

XhmikosR:
Perfect, thanks!

I hope we'll be able to remove lodash eventually, I've made quite a few patches so far, but it's not a simple task. Feel free to tackle any other instances and CC me.

BTW we might be able to remove other deps like prettysize.

XhmikosR:
I pushed a couple more changes, let me know if you have any objection :)

peterblazejewicz:
No, that's on your discretion here, thx!

XhmikosR:
After this PR lands, we should have 33 more instances. `isPlainObject` should be easy to replace too.

test/utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+            assert.equal(isFunction(/a/g), false);
+        });
+
+        it('should return false for a null value', () => {
+            assert.equal(isFunction(null), false);
+        });
+
+        it('should return false for an undefined value', () => {
+            assert.equal(isFunction(undefined), false);
+        });
```
Kreeg, line 34: + Hello! 
Please add negative tests for Object, Array, Number, String, Boolean

test/utils.test.js
```diff
@@ -0,0 +1,58 @@
...
+        it('should return true for a Function constructor', () => {
+            assert.equal(isObject(Function), true);
+        });
+
+        it('should return false for a null value', () => {
+            assert.equal(isObject(null), false);
+        });
+
+        it('should return false for an undefined value', () => {
+            assert.equal(isObject(undefined), false);
```
Kreeg, line 55: Same here, please provide additional negative tests for String, Boolean

lib/svg-sprite/utils/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+    return false;
+}
+
+/**
+ * Checks if value is the language type of Object. (e.g. arrays, functions, objects, regexes, new Number(0),
+ * and new String(''))
+ *
+ * @param {any} value The value to check.
+ * @return {boolean} Returns true if value is an object, else false.
```
Kreeg, line 20: Can you please add indentation for desription?

Kreeg (changes requested):
Hello! Thank you for this PR. Can you please do some minor changes I requested?

test/utils.test.js
peterblazejewicz: amended

test/utils.test.js
peterblazejewicz: amended, thx!

test/shape.test.js
```diff
@@ -5,10 +5,9 @@ const { readFileSync } = require('fs');
 const path = require('path');
 const { DOMParser } = require('@xmldom/xmldom');
 const SVGSpriter = require('../lib/svg-sprite.js');
+const { isObject } = require('../lib/svg-sprite/utils/index.js');
 const fixturesPath = require('./helpers/fixtures-path.js');
 
-const isObject = obj => typeof obj === 'object' && !Array.isArray(obj) && obj !== null;
```
XhmikosR, line 10: Makes me wonder if we should go with this solution?

There's also the `isPlainObject` we need to replace later, and I'm not sure of the difference between these 2 TBH.

lib/svg-sprite/utils/index.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Checks if value is the language type of Object. (e.g. arrays, functions, objects, regexes, new Number(0),
+ * and new String(''))
+ *
+ * @param {any} value The value to check.
+ * @returns {boolean} Returns true if value is an object, else false.
+ */
+function isObject(value) {
```
XhmikosR, line 22: Let's go with this, I think it's safer:

```js
function isObject(value) {
  return typeof value === 'object' && !Array.isArray(value) && value !== null;
}
```

XhmikosR:
I changed `isObject` to the stricter implementation but it seems 2 tests fail. I also added an `isPlainObject` utility but we need tests.

We can revert the last patches if you guys prefer to tackle these later, but I think the `isObject` changes are correct.

lib/svg-sprite/utils/index.js
peterblazejewicz: this change could break existing code (outcome of `writeFiles` is altered), including lib tests failing with a change. The checks does not work for arrays now for example. It's very terse and different from original lodash implementation

XhmikosR: Weird, that's what I've been using for years myself with vanilla JS. I still don't think `isObject` should return true for Array or Function, but I haven't checked the lodash implementation.

XhmikosR: OK, I had a closer look at the test failures and I think we should fix our tests instead. It seems we are passing an Array in writeFiles in some cases which is causing the test failures, otherwise my suggestion should be OK. @Kreeg what's your take on this?

Kreeg: @XhmikosR it is better to change. I'll take a look

Kreeg: @XhmikosR I see you already land these changes. That's good

Kreeg (approved):

XhmikosR:
@Kreeg how about we always pass an Object in `writeFiles`?

Note that this PR with the patches I landed needs tests for `isPlainObjects`.

Kreeg:
@XhmikosR I personally don't like `writeFiles` implementation and naming. But it is an issue for another PR.
I will land tests in a hour

lib/svg-sprite/utils/index.js
```diff
@@ -0,0 +1,51 @@
...
+function isFunction(value) {
+    if (value && typeof value === 'function') {
+        return true;
+    }
+
+    return false;
+}
+
+/**
+ * Checks if value is the language type of Object. (e.g. arrays, functions, objects, regexes, new Number(0),
```
XhmikosR, line 16: BTW this might need rewording (or the description of `isPlainObject`)

XhmikosR (approved):

PR #591, opened 2022-03-04
Use for...of loops when possible

XhmikosR:
For...of is more consistent and this way we don't need to explicitly check for own properties.

That being said, I hit a case where an error was thrown so I had to add a fallback, which means we could probably improve that part: https://github.com/svg-sprite/svg-sprite/pull/591/files#diff-79d35363069d02f38101ba80dad33c3dd32e39f32af88131c3365a3001a14b47R186

```js
for (const [extension, value] of Object.entries(this.config.render || {})) {
    tasks.push(createResourceTask(value, this.data, this._spriter, extension));
}
```

Also, unfortunately, the current tests don't cover everything yet.

TODO:

- [ ] Confirm all changes are good

### DIFF
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -201,7 +201,8 @@ if (typeof config.shape.transform === 'string') {
     const transform = String(config.shape.transform).trim();
     config.shape.transform = [];
     (transform.length ? transform.split(',').map(trans => String(trans).trim()) : [])
-        .forEach(function(transform) {
+        // TODO
+        .forEach(function(transform) { // eslint-disable-line unicorn/no-array-for-each
             if (transform.length) {
                 if (`shape-transform-${transform}` in argv) {
                     try {
@@ -218,6 +219,8 @@ if (typeof config.shape.transform === 'string') {
 }
 
 // Run through all sprite modes
+// TODO
+// eslint-disable-next-line unicorn/no-array-for-each
 ['css', 'view', 'defs', 'symbol', 'stack'].forEach(function(mode) {
     if (!argv[mode] && !(mode in JSONConfig.mode)) {
         delete this[mode];
@@ -225,6 +228,8 @@ if (typeof config.shape.transform === 'string') {
     }
 
     // Remove excessive render types
+    // TODO
+    // eslint-disable-next-line unicorn/no-array-for-each
     ['css', 'scss', 'less', 'styl'].forEach(function(render) {
         const arg = `${mode}-render-${render}`;
         if (render in this && !argv[arg] && (!(mode in JSONConfig.mode) || !('render' in JSONConfig.mode[mode]) || !(render in JSONConfig.mode[mode].render))) {
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -269,6 +269,7 @@ const files = argv._.reduce((f, g) => [...f, ...glob.sync(g)], []);
 
 for (let file of files) {
     let basename = file;
+    // TODO: get rid of variable redefinition
     file = path.resolve(file);
     const stat = fs.lstatSync(file);
     if (stat.isSymbolicLink()) {
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -75,8 +75,8 @@ function addOption(name, option) {
     }
 
     const { description, alias: optAlias, default: optDefault, map, ...children } = option;
-    for (const sub in children) {
-        addOption(`${name}-${sub}`, children[sub]);
+    for (const [key, value] of Object.entries(children)) {
+        addOption(`${name}-${key}`, value);
     }
 }
 
@@ -126,8 +126,8 @@ function writeFiles(files) {
 // Get document, or throw exception on error
 try {
     const options = yaml.load(fs.readFileSync(path.resolve(__dirname, 'config.yaml'), 'utf8'));
-    for (const name in options) {
-        addOption(name, options[name]);
+    for (const [key, value] of Object.entries(options)) {
+        addOption(key, value);
     }
 } catch (error) {
     console.log(error);
@@ -136,12 +136,12 @@ try {
 const { argv } = yargs;
 
 // Map all arguments to a global configuration object
-for (const map in optionsMap) {
-    if (!(optionsMap[map] in argv)) {
+for (const [key, value] of Object.entries(optionsMap)) {
+    if (!(value in argv)) {
         continue;
     }
 
-    addConfigMap(config, map.split('.'), argv[optionsMap[map]]);
+    addConfigMap(config, key.split('.'), argv[value]);
 }
 
 // Load external JSON config file
@@ -238,7 +238,7 @@ if (typeof config.shape.transform === 'string') {
 }, config.mode);
 
 // Remove excessive example options
-for (const mode in config.mode) {
+for (const mode of Object.keys(config.mode)) {
     const example = `${mode}-example`;
     if (!argv[example] && (!(mode in JSONConfig.mode) || !('example' in JSONConfig.mode[mode])) && 'example' in config.mode[mode]) {
         delete config.mode[mode].example;
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -108,9 +108,7 @@ function addConfigMap(store, path, value) {
  */
 function writeFiles(files) {
     let written = 0;
-    for (const key in files) {
-        const file = files[key];
-
+    for (const file of Object.values(files)) {
         if (isObject(file)) {
             if (file.constructor === File) {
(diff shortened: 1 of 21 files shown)


lib/svg-sprite.js
```diff
@@ -366,20 +366,20 @@ SVGSpriter.prototype._logStats = function(files) {
...
+            const file = resource.relative;
+            const ext = path.extname(resource.path).toUpperCase();
             exts[ext] = (exts[ext] || 0) + 1;
-            sizes[file] = pretty(files[mode][resource].contents.length);
+            sizes[file] = pretty(resource.contents.length);
         }
     }
 
-    this.info('Created %s', Object.entries(exts).map(ext => `${ext[1]} x ${ext[0].substr(1)}`).join(', '));
+    this.info('Created %s', Object.entries(exts).sort().map(ext => `${ext[1]} x ${ext[0].substr(1)}`).join(', '));
```
XhmikosR, line 378: I added this for consistency, but happy to split/drop the change.

lib/svg-sprite/mode/base.js
```diff
@@ -182,8 +182,8 @@ SVGSpriteBase.prototype._buildCSSResources = function(files, cb) {
         };
     };
 
-    for (const extension in this.config.render) {
-        tasks.push(createResourceTask(this.config.render[extension], this.data, this._spriter, extension));
+    for (const [extension, value] of Object.entries(this.config.render || {})) {
```
XhmikosR, line 185: This is the fallback I needed to add. Maybe we could make this branch more explicit with an if?

lib/svg-sprite/mode/symbol.js
```diff
@@ -137,9 +137,10 @@ SVGSpriteSymbol.prototype.layout = function(files, cb) {
                         removeAttributes.push(attribute.name);
```
XhmikosR, line 137: There's one `Array.prototype.forEach.call` here but I couldn't get rid of it last time I tried, and we don't have tests for this mode yet.

lib/svg-sprite/shape.js
```diff
@@ -730,17 +731,17 @@ SVGShape.prototype.setNamespace = function(ns) {
...
+            for (const elem of select('//*[@class]', this.dom)) {
                 const elemClassnames = [];
-                elem.getAttribute('class').split(' ')
-                    .filter(classname => classname.trim())
-                    .forEach(classname => {
-                        const substClassname = ns + classname;
-                        substClassnames[`.${classname}`] = substClassname;
-                        elemClassnames.push(substClassname);
-                    });
+                for (const classname of elem.getAttribute('class').split(' ')
```
XhmikosR, line 736: We could/should move these kind of long expressions to variables before the loop

lib/svg-sprite/mode/standalone.js
```diff
@@ -64,42 +64,44 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
...
-                    }] : [{
-                        expression: format(this.config.dimensions, shape.base),
-                        raw: format(this.config.dimensions, shape.base),
-                        first: true,
-                        last: true
-                    }]
-                }
-            };
+        if (shape.master) {
+            continue;
```
XhmikosR, line 68: not sure about this patch, needs more eyes.

lib/svg-sprite/mode/standalone.js
```diff
@@ -64,42 +64,44 @@ SVGSpriteStandalone.prototype._layout = function(files, cb, extend) {
...
+                    first: true,
+                    last: true
+                }]
+            }
+        };
 
-            this.data.shapes[index].__defineSetter__('svg', function(svg) {
-                this._svg = svg;
-            });
+        Object.defineProperty(this.data.shapes[index], '_svg', {
```
Kreeg, line 95: Maybe this changes should be in another PR?

XhmikosR: I didn't change that, make sure you have whitespace disabled

XhmikosR:
@Kreeg I'd like to land this if possible, could you review it please? I'll manually squash any patches before the merge.

XhmikosR:
Oh, and if you could fix the test failure, that would help!

I'm mostly not sure about the CLI changes since those are not covered by tests at all.

XhmikosR:
@Kreeg can you review this please? Tests pass now, so the thing that's not really tested is only the CLI.

lib/svg-sprite/mode/symbol.js
```diff
@@ -134,11 +134,11 @@ SVGSpriteSymbol.prototype.layout = function(files, cb) {
                     shapeDOM.localName = 'symbol';
                     const removeAttributes = [];
 
-                    Array.prototype.forEach.call(shapeDOM.attributes, attribute => {
-                        if (!symbolAttributes.has(attribute.name)) {
-                            removeAttributes.push(attribute.name);
+                    for (const { name } of Object.values(shapeDOM.attributes)) {
```
XhmikosR, line 137: This needs a second pair of eyes

Kreeg:
@XhmikosR sorry, I missed the notifications. I'll review this later this week.

bin/svg-sprite.js
```diff
@@ -220,13 +219,17 @@ if (typeof config.shape.transform === 'string') {
 }
 
 // Run through all sprite modes
+// TODO
+// eslint-disable-next-line unicorn/no-array-for-each
 ['css', 'view', 'defs', 'symbol', 'stack'].forEach(function(mode) {
```
Kreeg, line 224: This should also moved to constants in future

lib/svg-sprite.js
```diff
@@ -302,10 +303,10 @@ SVGSpriter.prototype._compile = function() {
 SVGSpriter.prototype._indexNamespace = function(index) {
     let ns = '';
 
-    for (let n = 0; n < this._namespacePow.length; n++) {
-        const c = Math.floor(index / this._namespacePow[n]);
+    for (const n of Object.values(this._namespacePow)) {
```
Kreeg, line 306: This is also already an array

lib/svg-sprite.js
```diff
@@ -383,20 +384,20 @@ SVGSpriter.prototype._logStats = function(files) {
...
+            sizes[file] = pretty(resource.contents.length);
         }
     }
 
     this.info('Created %s', Object.entries(exts).map(ext => `${ext[1]} x ${ext[0].substr(1)}`).join(', '));
 
-    Object.keys(sizes).sort().forEach(file => {
-        this.verbose('Created %s: %s', file, sizes[file]);
-    });
+    for (const [file, size] of Object.entries(sizes).sort()) {
```
Kreeg, line 398: The sorting here is changed. But also I'm not sure of sorting here is needed

bin/svg-sprite.js
XhmikosR: Yeah, this and the other TODO would be nice to switch to for of, but I couldn't make it work myself. Feel free to push to this branch.

XhmikosR:
@Kreeg feel free to push to this branch if you have any fixes. It'd help land this. Remember that I started the branch a long time ago, and since then quite a few things have changed,.

lib/svg-sprite.js
XhmikosR: Not sure either now that you mention. We could remove sorting. Feel free to push to this branch.

lib/svg-sprite/mode/symbol.js
Kreeg: LGTM

XhmikosR:
@Kreeg can we move some of the non-related patches to another branch? I personally have moved some of mine to https://github.com/svg-sprite/svg-sprite/compare/for-of...xmr/dev?w=1 so that we keep this branch only for the for...of changes.

XhmikosR:
I finished moving our non-related patches to the [xmr/dev](https://github.com/svg-sprite/svg-sprite/compare/for-of...xmr/dev?w=1) branch so that we land this one squashed and the rest as individual patches :)

lib/svg-sprite.js
XhmikosR: Alternatively, we could try using something like:

```js
for (const [file, size] of Object.fromEntries(Object.entries(sizes).sort())) {
    this.verbose('Created %s: %s', file, size);
}
```

XhmikosR:
@Kreeg LGTY so that we land this?

Kreeg:
@XhmikosR wait pls, this is a big PR, I checked only 60%

Kreeg (approved):

Kreeg:
@XhmikosR pls check my latest changes and we can go

XhmikosR:
@Kreeg thanks! I moved the logic into a variable. Assuming sorting is OK, we should be good to go. Worst case, we break the log sorting (which we could probably just drop anyway).

PR #595, opened 2022-03-04
feat: adding jsdoc linting

Kreeg:


package.json
```diff
@@ -97,6 +104,14 @@
       ],
       "eslint-comments/disable-enable-pair": "off",
       "guard-for-in": "off",
+      "jsdoc/check-values": [
+        "warn",
+        {
+          "allowedLicenses": ["MIT https://github.com/svg-sprite/svg-sprite/blob/main/LICENSE"]
```
Kreeg, line 110: there is 2 options here
1. keep this
2. change license to "MIT" (see https://spdx.org/licenses/)

package.json
```diff
@@ -97,6 +104,14 @@
...
+        }
+      ],
+      "jsdoc/no-undefined-types": ["error", {"definedTypes": ["SVGSpriter", "File", "SVGShape", "SVGSprite", "puppeteer", "HTMLElement"]}],
```
Kreeg, line 113: Or we can provide @typedef

lib/svg-sprite.js
```diff
@@ -30,7 +29,8 @@ const CPU_COUNT = os.cpus() && os.cpus().length;
 /**
  * SVGSpriter class
  *
- * @param {Object} config                Configuration
+ * @param {object} config                Configuration
+ * @class
```
Kreeg, line 33: Not sure about this

XhmikosR:
Not a huge fan of this myself but if it helps you, go ahead. That being said, a warning might be moot since the build won't fail. On the other hand, failing the build for this stuff might be too much.

PS. Unless you CC me I don't get notified. I have too many projects :/

lib/svg-sprite.js
XhmikosR: It's not a Class AFAICT. I tried in the past to switch to Classes but didn't succeed so I kept the prototype methods.

XhmikosR:
Nitpicking: indentation seems off in many places, not sure if there's any way to automate this too :)

Kreeg:
@XhmikosR indentation for this thing is available with prettier-jsdoc-plugin, but for now I'll eyeball it

PR #596, opened 2022-03-04
Downgrade csso to v4.

XhmikosR:
It's the same version used by svgo so this reduces the deps

@Kreeg just a reminder so that you don't accidentally merge a dependabot PR in the future unless svgo is updated to use csso@5 too :)

I'm still not sure if this works properly but anyway, the less packages, the better.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -178,9 +178,9 @@
       }
     },
     "node_modules/@eslint/eslintrc": {
-      "version": "1.1.0",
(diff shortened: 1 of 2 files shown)


Kreeg:
@XhmikosR OK! Is there any config to exclude some packages from dependabot?

Kreeg (approved):

XhmikosR:
Yeah, but we can ignore it through a comment in the PR.

PR #601, opened 2022-03-05
feat: test refactoring

Kreeg:
I don't like the result but it's just a step to split enormous test file


test/helpers/compare-svg-2-png-helper.js
```diff
@@ -0,0 +1,42 @@
...
+ * @param {Function} done             Callback
+ */
+module.exports = async(svg, png, expected, diff, done) => {
+    fs.mkdirSync(path.dirname(png), { recursive: true });
+    let browser;
+
+    try {
+        browser = await puppeteer.launch();
+        await convertSvg2Png(svg, png, browser);
+        looksSame.createDiff({
```
XhmikosR, line 25: BTW perhaps we could change this to create the diff file only when an error occurs?

Kreeg: @XhmikosR that's a good suggestion! OK

XhmikosR (approved):

PR #602, opened 2022-03-05
shape.js: use a Set for classnames

Kreeg:


lib/svg-sprite/shape.js
```diff
@@ -861,7 +861,7 @@ SVGShape.prototype._replaceIdAndClassnameReferencesInCssSelectors = function(str
 
             // Substitute class names within the selector
             if (classnames.length) {
-                classnames = [...new Set(classnames)]
+                [...new Set(classnames)]
```
XhmikosR, line 864: Unsure if this is covered right now.

Kreeg: @XhmikosR This does not need to cover. forEach doesn't return value. I saw you changed It from this 
```
// Substitute class names within the selector
	if (classnames.length) {
		classnames      = [...new Set(classnames)];
		classnames.sort(function(a, b){
                  return b.length - a.length;
              });
              classnames.forEach(function(classname) {
                  selText     = selText.split('.' + classname).join('.' + substClassnames['.' + classname]);
              }, this);
	}
```

XhmikosR: I know that forEach doesn't return anything, which is why I wonder if the code was right in the first place.

Not sure I follow your second sentence. When/where did I change that?

XhmikosR: If we have tests for class replacement, let me know. I don't remember right now.

Kreeg: @XhmikosR it was here https://github.com/svg-sprite/svg-sprite/commit/dd65765ffe103a49185d6de0f39dac918a2f898f/#diff-4065bb0d9fd86b292b781964d9713af8d9b891f6aa70470c236ba2a26cdf8459L839
And no, we don't have tests for class replacements. It can be done with snapshots in jest I think.

XhmikosR: Hmm, could be a mistake.

Keep in mind the patch you are linking was a big one and it was hanging around for a long time, so mistakes could have happened.

That being said, if we aren't 100% sure, let's convert the PR to a draft until we have covered this part of the code.

Kreeg: What I am saying here is that this code will function as it was. There was no error in current code, this is just removal of obsolete variable assignment.
I'm OK with converting this to draft.

XhmikosR: If you are confident, feel free to land it. It's just that I'd prefer it if we had a test beforehand :)

XhmikosR:
@Kreeg now might be a good time to revisit this maybe? I'm just not sure if we should overwrite `classnames` with a unique Set rather than just drop the assignment.

We need to revisit the patch that introduced the change to see what it meant to do and assuming it's covered by tests we should be safer :)

lib/svg-sprite/shape.js
```diff
@@ -835,10 +835,10 @@ SVGShape.prototype._replaceIdAndClassnameReferencesInCssSelectors = function(str
             const csssel = new CssSelectorParser();
             let sel = csssel.parse(selText);
             const ids = [];
```
XhmikosR, line 837: Not related to this PR, I just saw it here, we could probably use a Set for `ids` and possibly in other places where we expect unique data.

XhmikosR (approved):

PR #603, opened 2022-03-06
Remove rimraf devDependency

XhmikosR:


test/svg-sprite.test.js
```diff
@@ -63,10 +67,8 @@ async function compareSvg2Png(svg, png, expected, diff, done, msg) {
     }
 }
 
-before(done => {
-    rimraf(path.join(__dirname, '../tmp'), () => {
-        done();
-    });
+before(() => {
+    rm(path.join(__dirname, '../tmp'), { force: true, recursive: true });
```
Kreeg, line 71: It should be switched to sync function, or `done` callback should be provided, of should be switched to async/await.
I prefer callback

test/svg-sprite.test.js
```diff
@@ -38,6 +37,11 @@ const expectationsPath = require('./helpers/expectations-path.js');
 
 const dest = tmpPath;
 
+// This is needed so that we don't hit the `fs.rmdir`
+// deprecation warnings on Node.js >= 14.14.0
+// TODO Drop this when we drop support for Node.js 12
+const rm = fs.promises.rm ? fs.promises.rm : fs.promises.rmdir;
```
Kreeg, line 43: This can be simplified to `fs.promises.rm || fs.promises.rmdir;`

XhmikosR:
@Kreeg Feel free to push your suggestions and I'll check them out later.

test/svg-sprite.test.js
XhmikosR: I tend to prefer the plain old ternary since we can't use `??` yet, but both achieve the same. Generally, nowadays, I don't like using short and seemingly "smarter" code, but rather more explicit :)

PR #605, opened 2022-03-08
feat: final tests refactoring

Kreeg:


test/svg-sprite.test.js
```diff
@@ -1,382 +0,0 @@
-'use strict';
-
-/**
- * svg-sprite is a Node.js module for creating SVG sprites
- *
- * @see https://github.com/svg-sprite/svg-sprite
- * @author Joschi Kuphal <joschi@kuphal.net> (https://github.com/jkphl)
- * @copyright © 2018 Joschi Kuphal
- * @license MIT https://github.com/svg-sprite/svg-sprite/blob/main/LICENSE
```
Kreeg, line 9: @XhmikosR @jkphl mind that this was removed now. Is it OK? If not - we need to copy that to all the test files

test/helpers/remove-temp-path.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const fs = require('fs');
+const tmpPath = require('./tmp-path.js');
```
XhmikosR, line 4: How about having a constants file with all these test helpers one-liners? Just thinking out loud since it looks like unnecessary abstraction :)

Kreeg: Yeah that'll be better.

test/helpers/remove-temp-path.js
```diff
@@ -1,13 +1,13 @@
 'use strict';
 
 const fs = require('fs');
-const tmpPath = require('./tmp-path.js');
+const { paths } = require('../constants.js');
```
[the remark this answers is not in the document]
Kreeg: @XhmikosR I think no. Because they do different things. But i move constants.js into helpers folder

XhmikosR (approved):

PR #611, opened 2022-03-19
Add prettier to xo config and switch to two-space indentation

Kreeg:
@XhmikosR for me, conversations in commits are quite inconvenient. Let's communicate here.
Have you seen this changes? Do you agree/disagree with them?
Sorry in advance for this PR.


XhmikosR:
So, here we have 2 choices:

1. try to stay close to the xo style we are already using
2. go with prettier's defaults and adapt xo

In the first case, not sure what the gain is, though. Just being able to format other files except for js/ts etc?

I don't mind going with the second way, as long as we agree on a common set of options.

Kreeg:
@XhmikosR let's go with second variant. Then step by step, file by file we will include prettier to all code base.

Kreeg:
@XhmikosR should we land it?

XhmikosR:
Please hold on to this. I haven't had a chance to look at it yet.

bin/svg-sprite.js
```diff
@@ -3,17 +3,18 @@
 'use strict';
 
 /**
- * svg-sprite is a Node.js module for creating SVG sprites
+ * Svg-sprite is a Node.js module for creating SVG sprites
```
XhmikosR, line 6: Is there any way we don't touch the license headers?

XhmikosR:
So far I see the following:

1. I don't like prettier using single quotes in yaml, but I can live with it
2. I don't like the license header changes
3. I definitely don't like prettier's HTML handling
4. prettier chokes on our templates
5. format check is not enforced on CI; we should probably add it there too if we go with this approach
6. I switched to two-space indentation because it feels more natural, but it's hard to see the diff because it's huge

bin/svg-sprite.js
Kreeg: Yes, https://github.com/hosseinmd/prettier-plugin-jsdoc/#Options
`jsdocCapitalizeDescription: false`

XhmikosR:
Closing for now. I landed the 2-space indentation in #812

PR #614, opened 2022-03-19
CI: add macOS tests too

XhmikosR:
Not sure if we want to do this... We should test it, but it increases CI time. Basically macOS and Windows are a lot slower than Ubuntu on GitHub Actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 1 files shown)


Kreeg:
I think this is too much. We don't need this now. 
Maybe it can be set as option in actions to run some PR's check on macOS too?

XhmikosR:
Not sure how you mean exactly. We could run all Node.js versions on Ubuntu and just the minimum one on the other two OS'es.

BTW we have duplicate CI runs and it's a known thing. I'm just used to have CI run without PRs too, but we can change that.

Issue #604, opened 2022-03-08
Remove/replace lodash dependency


XhmikosR:
Opening this separately since it requires more work (refs #598)

Right now on main we have 8 remaining lodash functions:

- [x] `cloneDeep` (see #622)
- [x] `escape` (see #620)
- [x] `invoke` (see #461)
- [x] `merge` (see #619)
- [x] `pick` (see #463)
- [ ] `trim` (see #418)
- [ ] `trimStart` - wondering if we can use `foo.replace()`? We might need to escape any special regex characters, though
- [x] `zipObject` (see #616)

PR #616, opened 2022-03-20
Replace `_.zipObject` with vanilla JS

XhmikosR:
TODO:

- [x] decide with which solution to go (1st or 2nd patch); I haven't done any benchmarks, so any help to decide is welcome
- [x] add more tests (help welcome!)

Refs #604


test/utils.test.js
```diff
@@ -196,4 +202,14 @@ describe('utils', () => {
             expect(isPlainObject(Symbol('test'))).toBe(false);
         });
     });
+
+    describe('zipObject', () => {
+        it('should return the zipped object', () => {
+            expect(zipObject(['a', 'b'], [1, 2])).toStrictEqual({ a: 1, b: 2 });
```
Kreeg, line 208: Need more tests to cover possible variants, like 
+`(['a', 'b', 'c'], [1, 2])`
+`(['a'], [1, 2])`

lib/svg-sprite/utils/index.js
```diff
@@ -41,9 +41,19 @@ function isString(value) {
     return Object.prototype.toString.call(value) === '[object String]';
 }
 
+/**
+ * @param {Array} array1    First array
+ * @param {Array} array2    Second array
+ * @returns {object}        The zipped Object
+ */
+function zipObject(array1, array2) {
```
Kreeg, line 49: I'm thinking this arguments should be checked to be an array. Instead throw error.

test/utils.test.js
XhmikosR: Feel free to add them but I'm not sure if they'll work. What did the lodash zipObject return in these cases?

lib/svg-sprite/utils/index.js
XhmikosR: I thought about this too, but didn't add this yet. Feel free to push :)

Kreeg:
@XhmikosR I vote for second patch. I think It's just clearer.

test/utils.test.js
XhmikosR: @Kreeg did you confirm what lodash is returning with unbalanced arrays?

XhmikosR: So, the current behavior is the same, right?

Kreeg: @XhmikosR yes

PR #620, opened 2022-03-20
Replace `_.escape` with native JS

XhmikosR:
Refs #604

- [x] confirm this is the same as the lodash method
- [x] see if we can use `entityMap`'s keys
- [x] add more tests?
- [x] should not throw should use `not.throw`

An alternative solution would be to go with a separate (small) package, which is an acceptable solution too, assuming we finish with all other lodash instances.

### DIFF
--- a/lib/svg-sprite/sprite.js
+++ b/lib/svg-sprite/sprite.js
@@ -10,9 +10,8 @@
  */
 
 const { Buffer } = require('buffer');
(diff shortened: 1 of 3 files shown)


Kreeg:
@XhmikosR OK, there is no problem to squash them, I just though that it would be nice to keep commits from different authors.  
As I think the current realization is decent.

PR #621, opened 2022-03-20
add test for stack mode

Kreeg:


Kreeg:
I'm sure we can drop 2048x2048 svg I've added for calculations tests. It causes some glitches in png references (img is way bigger than it's parent. Or we can adjust css to deal with such images and update all expected references.

test/tmpl/stack.html
```diff
@@ -0,0 +1,78 @@
...
+                padding: 0;
+                text-align: center;
+                display: inline-block;
+                line-height: 1;
+                vertical-align: middle;
+            }
+            figcaption {
+                display: none;
+            }
+            i {
```
XhmikosR, line 47: The tmpl doesn't have any `i` elements. Should the test have them?

XhmikosR:
> I'm sure we can drop 2048x2048 svg I've added for calculations tests. It causes some glitches in png references (img is way bigger than it's parent. Or we can adjust css to deal with such images and update all expected references.

As long as we still test things out, you can make a PR and we discuss it :)

test/tmpl/stack.html
Kreeg: I think no. I'll remove it

PR #619, opened 2022-03-20
Switch to the `lodash.merge` package

XhmikosR:
Refs #604 

We should probably try and land #463 and #616 first.

PR #622, opened 2022-03-20
Replace `_.cloneDeep` with `_.merge`

XhmikosR:
@Kreeg It's not more readable, but feel free to push any changes to improve readability.

Split from #619 since that should probably land after we are done with all other lodash instances.


Kreeg:
Yeah I think that should be placed in `utils.js` and called like `cloneDeep`

XhmikosR:
I think it's OK to keep it like this for now. We can revisit later.

Kreeg (approved):

PR #623, opened 2022-03-20
Rename npm scripts

XhmikosR:
We could also just drop them since one can do `npx jest --foo` or `npm run jest -- --foo`, but I guess it doesn't hurt.


Kreeg:
I'd rather to stay with script names instead of npx

PR #624, opened 2022-03-20
CI: limit pushes only on main

XhmikosR:


Kreeg:
Basically, it will run CI only on push in main branch?

XhmikosR:
Yeah, and PRs.

Kreeg:
Let's try and see what will happen

PR #627, opened 2022-03-21
fix: checking passed config.log to be instance of winston.Logger

### LINKED ISSUE #626: Passing winston logger to config causes Error
```
const log = winston.createLogger();
const config = new SVGSpriterConfig({
    log,
});
```
throws
```
Uncaught TypeError: Function has non-object prototype 'undefined' in instanceof check
    at Function.[Symbol.hasInstance] (<anonymous>)
```
Because `winston.Logger` has no constructor

Kreeg:
Fixes #626

### DIFF
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -13,7 +13,6 @@ const fs = require('fs');
 const path = require('path');
 const yaml = require('js-yaml');
 const winston = require('winston');
-const Logger = require('winston/lib/winston/logger.js');
 const { isFunction, isObject, isPlainObject, isString } = require('./utils/index.js');
 
 /**
@@ -87,6 +86,19 @@ const defaultSVGConfig = {
 };
 
 /**
+ * @param {any} logger  probable logger
+ * @returns {boolean}   result
+ */
+const isWinstonLogger = logger => {
+    return (
+        isObject(logger) &&
+        logger.level !== undefined &&
+        Array.isArray(logger.transports) &&
+        isFunction(logger.log)
+    );
+};
+
+/**
  * SVGSpriter configuration
  *
  * @param {object} config                 Configuration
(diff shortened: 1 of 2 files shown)


lib/svg-sprite/config.js
```diff
@@ -13,12 +13,13 @@ const fs = require('fs');
 const path = require('path');
 const yaml = require('js-yaml');
 const winston = require('winston');
+const Logger = require('winston/lib/winston/logger.js');
```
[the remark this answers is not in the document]
Kreeg: No. But official exports has no Logger constructor. This behavior changed between v2 and v3 of package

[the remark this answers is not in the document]
XhmikosR: Hmm, unsure how proceed. It's just that this is fragile. Isn't there any other way to check for this?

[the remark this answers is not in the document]
Kreeg: I prefer to drop this checking of being instance of winston.Logger. Because the other part of `if` statement is checking of log is winston Logger but not an actual instance of `Logger`

lib/svg-sprite/config.js
```diff
@@ -85,6 +85,19 @@ const defaultSVGConfig = {
...
 
+/**
+ * @param {any} logger  probable logger
+ * @returns {boolean}   result
+ */
+const isWinstonLogger = logger => {
+    return (
+        isObject(logger) &&
+        logger.level !== undefined &&
+        Array.isArray(logger.transports) &&
```
Kreeg, line 96: @XhmikosR here is the thing.
`lodash.isObject` returns true if array is passed. Our util is not. So there was an error here too because `transports` is an array

XhmikosR: Ouch... We might need to revisit our util then.

XhmikosR: BTW I prefer our isObject because it's tighter, but since we might have broken some stuff, better add a comment and accept Arrays too for now. We should revisit it later if needed.

Kreeg: Ok. I started to write tests. The goal is to cover entire project. This will find such errors. Our `isObject` is better, I agree

PR #631, opened 2022-03-22
add more layouter tests

Kreeg:


XhmikosR (approved):

XhmikosR:
BTW @Kreeg we can remove `bin` from coverage until we have tests if it helps, I just felt we were giving the wrong impression and I hadn't noticed that some files weren't covered at all. That's why I explicitly included bin and lib folders.

Kreeg:
@XhmikosR I think `bin` being in coverage is ok.

PR #632, opened 2022-03-23
Convert SVGSpriteLayouter to ES6 class

Kreeg:
It's safe since this file is totally covered by test

### DIFF
--- a/lib/svg-sprite/layouter.js
+++ b/lib/svg-sprite/layouter.js
@@ -59,7 +59,8 @@ const defaultConfig = {
         bust: false
     }
 };
-const defaultVariables = {
+
+const defaultMustacheVariables = {
     date: new Date().toGMTString(),
     invert() {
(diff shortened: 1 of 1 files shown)


XhmikosR:
Do you plan to convert everything to Classes? I can't access if that will be a breaking change or not yet :)

XhmikosR:
I quickly made https://github.com/svg-sprite/svg-sprite/compare/queue-class?w=1, feel free to use it later if you think we could switch everything to classes. Otherwise, I'm not sure it'll make sense to change only a few files :)

Kreeg:
@XhmikosR yea I plan but with step by step process. The tests came first and covers all functionality. Then refactoring. If something is breaking in tests - this is a breaking change. Otherwise it's safe.
As for the Queue to es6 - it's important to cover queue.js by tests first

PR #640, opened 2022-03-26
Convert Queue to an ES6 class

XhmikosR:


lib/svg-sprite/queue.js
```diff
@@ -19,88 +19,76 @@ const Shape = require('./shape.js');
...
-    if (this._files.length && this.active < this._spriter._limit) {
-        ++this.active;
-        const file = this._files.shift();
-        let shape;
-        let spriter;
+    /**
+     * Try to process an item in the queue
+     */
+    process() {
+        if (this._files.length && this.active < this._spriter._limit) {
```
XhmikosR, line 51: Maybe it makes sense to return early here? Feel free to push any patches to this branch.

PR #644, opened 2022-03-26
Convert Config to an ES6 class

XhmikosR:
10 more files left, I'll leave those to you @Kreeg :)


Kreeg:
@XhmikosR cool! But let's wait till all lines will be covered, since the work on this is already started. I'll merge this immediately after :)

PR #647, opened 2022-03-27
add tests for `transform/svgo`

Kreeg:


test/svg-sprite/transform/svgo.test.js
```diff
@@ -0,0 +1,212 @@
...
+
+const TEST_SVG = '<svg></svg>';
+
+jest.mock('svgo');
+
+describe('testing transforms.svgo', () => {
+    let noop;
+
+    beforeEach(() => {
+        noop = jest.fn();
```
XhmikosR, line 14: Isn't `noop` the same across the tests? If so, shouldn't we use the same instance/variable?

Kreeg: not it isn't. before each test it replaces with new instance of jest.fn. Additionally, all the mocks are cleared and reset after each test.

PR #653, opened 2022-04-02
Added tests for `SVGShape`

Kreeg:
small refactoring

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -148,6 +148,7 @@ function SVGShape(file, spriter) {
     this.svg = { current: this.source.contents.toString(), ready: null };
     this.name = path.basename(this.source.path);
     this.config = merge(merge({}, defaultConfig), this.spriter.config.shape || {});
+    // todo: get rid of setting empty object if this.spriter.config.shape is not provided (https://github.com/svg-sprite/svg-sprite/pull/653#discussion_r841069781)
 
     if (!isFunction(this.config.id.generator)) {
         this.config.id.generator = createIdGenerator(isString(this.config.id.generator) ? this.config.id.generator + (this.config.id.generator.includes('%s') ? '' : '%s') : '%s');
--- a/test/svg-sprite/shape/shape.constructor.test.js
+++ b/test/svg-sprite/shape/shape.constructor.test.js
@@ -1,0 +1,259 @@
+'use strict';
+
+/* eslint-disable max-nested-callbacks */
+
+const path = require('path');
+const SVGShape = require('../../../lib/svg-sprite/shape.js');
+
+describe('testing Shape.constructor', () => {
+    describe('testing constructor', () => {
+        const TEST_FILE = {
+            contents: '<svg>TEST CONTENT</svg>',
+            path: 'test_path',
+            relative: 'test_relative'
+        };
+
+        it('should set expected initial values', () => {
+            expect.hasAssertions();
+
+            const TEST_SPRITER = {
+                config: {
+                    shape: { meta: {}, align: {} },
+                    id: {}
+                },
+                verbose: jest.fn()
+            };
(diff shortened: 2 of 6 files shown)


lib/svg-sprite/shape.js
```diff
@@ -165,8 +165,8 @@ function SVGShape(file, spriter) {
 
     // Determine meta & alignment data
     const relative = path.basename(this.source.relative, '.svg');
-    this.meta = this.id in this.config.meta ? this.config.meta[this.id] : (relative in this.config.meta ? this.config.meta[relative] : {});
-    this.align = Object.entries({ ...this.config.align['*'], ...(this.id in this.config.align ? this.config.align[this.id] : (relative in this.config.align ? this.config.align[relative] : {})) });
+    this.meta = this.config.meta && this.id in this.config.meta ? this.config.meta[this.id] : (this.config.meta && relative in this.config.meta ? this.config.meta[relative] : {});
```
Kreeg, line 168: I added these changes because without them it's impossible to cover the file by the tests. The problem is if we do not provide `spriter.config.shape` , the `config.meta` and `config.align` became `undefined`. Because at the line 150 `config` merges with empty object. 
```
this.spriter.config.shape || {}
```

We should either keep this changes or remove this condition to merge with `this.spriter.config.shape` only

lib/svg-sprite/shape.js
```diff
@@ -369,7 +369,7 @@ SVGShape.prototype._initSVG = function() {
     this.dom = new DOMParser({
         locator: {},
         errorHandler(level, message) {
-            const e = new Error(format('Invalid SVG file (%s)', message.split('\n').join(' ')));
+            const e = new Error(format('Invalid SVG file (%s)', message.replace(/\s+/g, ' ').split('\n').join(' ')));
```
Kreeg, line 372: the `message` contains multiple spaces here. It appears normally in console, but it's hard to test.

lib/svg-sprite/shape.js
```diff
@@ -432,9 +432,9 @@ SVGShape.prototype.getDimensions = function() {
 /**
  * Set the dimensions of this shape
  *
- * @param {number} width        Width
- * @param {number} height       Height
- * @returns {SVGShape}          Self reference
+ * @param {number|string} width        Width
```
Kreeg, line 435: We should either keep this changes in docs or remove parseFloat

Kreeg:
Hi @XhmikosR Can you please check this? I know it adds more than just tests but I need this for the future tests

lib/svg-sprite/shape.js
XhmikosR: I'm not a fan of changing the code just for the tests. I wouldn't test this for the time being if it's causing issues.

lib/svg-sprite/shape.js
XhmikosR: Not sure what you mean exactly? If it's inconsistent with the docs, it might be a bug or a leftover.

lib/svg-sprite/shape.js
XhmikosR: Unsure how to proceed, I'll leave the choice to you. Just keep in mind that if it's something for tests only, we shouldn't change the code. If it's a real bug (which it seems so), then you can go ahead with a fix.

Kreeg: @XhmikosR OK, I'll reverse this code back and we'll just ignore this line for now. Not a big problem. We'll go back to this later, I'll add **TODO**

Kreeg:
@XhmikosR should I split it into 4 different PRs? Or is it OK to have such big one?

XhmikosR:
@Kreeg your call, ideally these should be at least separate commits and not squashed when merging.

Kreeg:
@XhmikosR I split this PR into different commits. Can you check, please?

PR #659, opened 2022-04-07
add more tests for `SVGShape`

Kreeg:


test/svg-sprite/shape/shape.svg.test.js
```diff
@@ -255,3 +255,121 @@ describe('testing setSVG()', () => {
...
+            shape.height = TEST_HEIGHT / 2;
+            shape.config.spacing.box = 'icon';
+            shape.config.dimension.maxHeight = TEST_HEIGHT;
+            shape.config.dimension.maxWidth = TEST_WIDTH;
+
+            shape._setDimensions(jest.fn());
+
+            expect(shape._scale).toBe(2);
+            expect(shape.height).toBe(100);
+            expect(shape.width).toBe(300); // 300 * 80 / 100
```
XhmikosR, line 350: the comment seems wrong here I think

XhmikosR: What do you think about using the actual expressions instead of the calculated result?

Kreeg: I originally wrote here the expressions. But rewrote to calculated results. I think this might be OK, we should test it

Kreeg: @XhmikosR it's much better now, thanks

PR #660, opened 2022-04-08
Update dependencies

XhmikosR:
* @xmldom/xmldom         ^0.8.1  →   ^0.8.2
* glob                   ^7.2.0  →   ^8.0.1
* winston                ^3.6.0  →   ^3.7.2
* eslint-plugin-jest    ^26.1.3  →  ^26.1.4
* eslint-plugin-jsdoc   ^38.1.4  →  ^38.1.6
* playwright-chromium   ^1.20.1  →  ^1.21.0
* sass                 ^1.49.11  →  ^1.50.0
* yargs                 ^17.4.0  →  ^17.4.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@resvg/resvg-js": "^1.4.0",
-        "@xmldom/xmldom": "^0.8.1",
+        "@xmldom/xmldom": "^0.8.2",
         "async": "^3.2.3",
         "css-selector-parser": "^1.4.1",
         "csso": "^4.2.0",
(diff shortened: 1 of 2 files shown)


XhmikosR:
@Kreeg So, eslint-plugin-jsdoc@39 drops support for Node.js 12. It doesn't affect our CI since we run lint on Node.js 16, though.

Unsure if we should just drop Node.js 12 at this point. Technically we can't because we are in the beta phase, but being that Node.js 12 is causing us issues with resvg and that it's going EOL soon we might need to reconsider.

Kreeg:
@Kreeg let's freeze the current version then, we'll rethink about this after we drop node 12 support

XhmikosR:
Maybe we should keep glob to v7 too. v8.x includes minimatch@5 which has a breaking change https://github.com/isaacs/minimatch/compare/v4.2.1...v5.0.1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

PR #665, opened 2022-04-19
add tests for `SVGSpriteBase`

Kreeg:


lib/svg-sprite/mode/base.js
```diff
@@ -135,7 +135,7 @@ SVGSpriteBase.prototype._initData = function(data) {
                 renderConfig.dest = path.resolve(this.config.dest, this.config.example.dest);
             }
         } else if (this.config.example !== true) {
-            renderConfig = false;
+            renderConfig = false; // todo: fix bug with this branch on line 142 (boolean has no "dest" property)
```
XhmikosR, line 138: Which branch are you referring to?

test/svg-sprite/mode/base.test.js
```diff
@@ -0,0 +1,438 @@
...
+                prefix: '1',
+                sprite: '',
+                example: {
+                    dest: '.'
+                }
+            };
+            const { cls } = getClassAndInitFn();
+            const base = new cls(TEST_SPRITER, TEST_CONFIG, {}, '');
+            const TEST_SVG = 'svg';
+            const TEST_SPRITE = 'sprite-ae8eb96d.svg';
```
XhmikosR, line 426: I'd just test for the number of characters here, like `^sprite-\w{8}\.svg$`. I prefer not using regex when possible but I think it makes sense here.

XhmikosR: BTW, unsure why we use 8 characters. I was under the impression 7 was more common.

lib/svg-sprite/mode/base.js
Kreeg: @XhmikosR current `else if` branch (line 138)

XhmikosR: Ah, I see. Indeed, good catch 👍

XhmikosR (approved):

PR #683, opened 2022-05-23
Update dependencies

XhmikosR:
@Kreeg there might be some new failures, haven't looked at them closely. Feel free to push here to fix any issues :)

EDIT: we can also disable any rules if needed.

### DIFF
--- a/example.js
+++ b/example.js
@@ -62,7 +62,7 @@ function addFixtureFiles(spriter, files) {
         spriter.add(
             path.resolve(path.join(cwd, file)),
             file,
-            fs.readFileSync(path.join(cwd, file), 'utf-8')
+            fs.readFileSync(path.join(cwd, file), 'utf8')
         );
     });
     return spriter;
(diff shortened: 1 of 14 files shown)


lib/svg-sprite.js
```diff
@@ -246,7 +246,7 @@ SVGSpriter.prototype._compile = function() {
         const [config, cb] = this._compileQueue.shift();
 
         // If this is a modeless run
-        if (config === {}) {
+        if (Object.keys(config).length === 0) {
```
Kreeg, line 249: Here was the error. Yes, this branch was never executed before. And the tests was kind of exploiting this bug. 
I changed the test, but TBH this should be in another PR.

Kreeg (approved):

lib/svg-sprite.js
XhmikosR: But this is now flagged by ESLint here so I'm not sure. I will probably rebase the PR and just split the patches before the merge.

Thanks for fixing the issue!

PR #688, opened 2022-05-28
Downgrade playwright due to different screenshot quality with & without plugged power source

Kreeg:
@XhmikosR new version brings back the old issue with different screenshot qualities with unplugged notebook. 
I fixed the version which is not causing this problem.
https://github.com/microsoft/playwright/issues/14460

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "jest": "^28.1.0",
         "less": "^4.1.2",
         "pixelmatch": "^5.3.0",
-        "playwright-chromium": "^1.22.2",
+        "playwright-chromium": "1.21.1",
         "pngjs": "^6.0.0",
         "sass": "^1.52.1",
         "stylus": "^0.57.0",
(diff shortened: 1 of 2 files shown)


XhmikosR:
What issue is that exactly?

You should report it upstream too and switch to a fixed version not a range.

XhmikosR:
Also, keep in mind we aren't going to stick to a version forever just for that. Eventually, either the issue needs to be fixed, or we find another solution.

XhmikosR:
And finally, try to make the commit messages/PRs more descriptive; in the future "fixing the previous playwright version" won't mean anything...

Kreeg:
@XhmikosR yeah sorry. All the regression tests are suddenly started to fail and I just spent 1 hour for debugging only to realize that this is a problem with new playwright's version. So I failed with PR naming, my bad :(
Yes, I know that we can stick with the the old version forever. I reported the issue to the upstream. This just took a bit of time. I should have created this PR after reporting the issue

XhmikosR:
Can you link to the upstream issue you reported too in the PR description please?

XhmikosR:
BTW I have also learned to live with tests spawning too many processes and crippling my Windows VM. I think this and my issue are the remaining ones that affect us. Not super-pressuring to fix, but would be nice if we could find a solution for both at some point in the future.

I also think we should proceed with releasing v2.0.0 stable soon but I'd need some help with updating the testing framework for the gulp and grunt wrappers (they still use svg2png which works but it's old). I'll ping you after I make PRs there.

Kreeg:
@XhmikosR I've already done it.

XhmikosR (approved):

Kreeg:
@XhmikosR Ok, cool. I thought we are far from 2.0.0 stable but there is so much new features and bug fixes comparing to 1.5.4 so this might be a good idea

XhmikosR:
I don't mind waiting if you want, I just think most of the consumers of the packages don't even know there's a v2.0.0 beta and there have been some good changes so far. Let me know what you think if you have another schedule on mind.

Kreeg:
@XhmikosR in ideal world this would be a moment when the all files are covered by unit tests. But the thing is I covered the easiest parts of the logic. I think the rest logic is the hardest and will take a lot of time. Consider that I am going to 2 week vacation the next week. All this things pushes back the "ideal" release. So lets go with the current changes and do a stable 2.0.0 release. You're right, the real consumers should not just wait for unit tests.

XhmikosR:
Cool, I'll add it to my schedule to release v2.0.0 stable in the next days and we'll see how it goes.

PS. I haven't forgotten about your prettier PR; I just need to find some time to adapt it to our needs.

Kreeg:
@XhmikosR that's OK, I just decided to update all my PRs before my vacation

XhmikosR:
@Kreeg does your issue still happen with the latest playwright?

Kreeg:
@XhmikosR yes. The problem is that I have not so much time to do all the tests to locate the problem (which is somewhere in chromium).

XhmikosR:
OK, we should try to update it at some point because it's already behind a lot.

PR #701, opened 2022-06-22
Update dependencies

XhmikosR:
* async                 ^3.2.3  →   ^3.2.4
* eslint-plugin-jest   ^26.4.6  →  ^26.5.3
* jest                 ^28.1.0  →  ^28.1.1
* less                  ^4.1.2  →   ^4.1.3
* sass                 ^1.52.1  →  ^1.52.3
* xo                   ^0.49.0  →  ^0.50.0

@Kreeg what is the plan with playwright-chromium? Do we still need to stick to the old version?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@resvg/resvg-js": "^2.0.1",
         "@xmldom/xmldom": "^0.8.2",
-        "async": "^3.2.3",
+        "async": "^3.2.4",
         "css-selector-parser": "^1.4.1",
         "csso": "^4.2.0",
         "cssom": "^0.5.0",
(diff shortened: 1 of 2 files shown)


Kreeg:
@XhmikosR I need to do 1 more test with playwright. I am aware of vulnerability in old version (found by npm audit). The problem is this test would take long to reproduce (I need to download each version of playwright and replace chromium version inside of each). I'll do it when i have a little bit more free time.  
For now we still need this old version to get more stable tests across multiple platforms. All the tests are failed on my old mackbook pro 2013 with the new playwright version, and passed with the old one. :(

XhmikosR:
OK, thanks for the reply. I will keep the old version for now, but we should figure something out at some point.

Issue #702, opened 2022-06-22
v2.0.0-beta6 escape issue in symbol mode


XhmikosR:
```diff
 bootstrap-icons.svg | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/bootstrap-icons.svg b/bootstrap-icons.svg
index 99df40fb4..be7b3ea38 100644
--- a/bootstrap-icons.svg
+++ b/bootstrap-icons.svg
@@ -1 +1 @@
-<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><symbol class="bi bi-123"...
\ No newline at end of file
+<svg xmlns="http:&#x2F;&#x2F;www.w3.org&#x2F;2000&#x2F;svg" xmlns:xlink="http:&#x2F;&#x2F;www.w3.org&#x2F;1999&#x2F;xlink"><symbol class="bi bi-123"
\ No newline at end of file

```

@Kreeg it seems we have a regression probably after #620.

To reproduce:

```
git clone https://github.com/twbs/icons.git -b deps
npm i
npm run icons-sprite
```

and see `./bootstrap-icons.svg`.

PR #704, opened 2022-06-22
Added snapshots testing for regression tests

Kreeg:
@XhmikosR this will check all svgs in regression tests to match the given shapshots. As mentioned here #702 it is necessary to prevent this to happen again.
But there is caveat we should deal with. This matching will fail if any character in generated svg is changed. This may happen after `svgo` or any other packages is updated. In this case all the snapshots must be updated with `npm run test -- -u`


Kreeg:
@XhmikosR this is odd since all the tests are passed. Maybe you tried it in old version of branch `node-14`?
Can you provide a stacktrace? I don't have any windows machine

XhmikosR:
Hmmm, wait now it works fine. Not sure what happened!

PR #721, opened 2022-08-03
fix(docs): fix example config for inline symbols

strager:
'inline' isn't a mode. 'inline' is a config property for the 'symbol'
mode. Fix the example in the docs to set the intended property.


Kreeg:
@strager thank you!

PR #726, opened 2022-08-27
Replaced `Error.errno` with Error classes

### LINKED ISSUE #645: Remove `Error.errno`
And probably Error.name and just throw an `Error` with the proper message

Kreeg:
Closes #645


XhmikosR (approved):

XhmikosR:
Maybe not much, but each new file adds up to the module load time since we aren't using ESM. Anyway, just keep that in mind for the future.

PR #742, opened 2022-10-26
refactor: drop lodash.trim dependency

strager:
Use JavaScript's built-in String.prototype.trim method instead of lodash.trim, reducing dependency bloat.

In one case, we were calling lodash.trim with a Buffer instance. lodash.trim calls toString on its argument, so we didn't notice this issue before. Continue calling toString before calling String.prototype.trim to preserve behavior.

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -14,7 +14,6 @@ const path = require('path');
 const events = require('events');
 const os = require('os');
 const { format } = require('util');
-const trim = require('lodash.trim');
 const trimStart = require('lodash.trimstart');
 const merge = require('lodash.merge');
 const File = require('vinyl');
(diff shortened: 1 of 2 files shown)


Kreeg (commented):
Hello! Thanks for the PR
Please update package-lock as well. Thanks

XhmikosR:
Thanks for the PR. Last time I tried this, it caused some issues, but my patch wasn't the same as yours and we didn't have as good coverage as we do now. So, I guess if tests pass we should be fine.

I'd still manually test gulp-svg-sprite and a few projects just to be safe, but I don't know when I'll have the time.

XhmikosR:
@strager please rebase your branch and enable the option to allow edit from maintainers so that we can edit it too.

XhmikosR:
@Kreeg LGTY? If so, I'll cherry pick it manually later.

XhmikosR:
NVM, I made #749 so that we can move on, thanks @strager!

PR #743, opened 2022-10-26
fix(docs): fix incorrect type annotation

strager:
SVGSpriter.prototype.add is called with Buffer instances in some tests, and that seems to be intentionally supported. Fix the outdated JSDoc which says that only strings are allowed.


Kreeg (approved):
@XhmikosR  I think we should narrow down expected types, but for now it's OK.

Issue #598, opened 2022-03-04
Reduce direct dependencies/devDependencies


XhmikosR:
With jest, we might be able to remove the following devDependencies:

1. c8
2. should
3. sinon

As for dependencies, we might be able to remove:

1. lodash
2. prettysize
3. any other deps?

chalkygames123:
I'd like to recommend [`fdir`](https://github.com/thecodrr/fdir) as an alternative to `glob`.

Authors post: https://dev.to/thecodrr/how-i-wrote-the-fastest-directory-crawler-ever-3p9c

Kreeg:
Seems like a good solution. It is lightweight, dependency-free, I'm down with that.

XhmikosR:
One thing to keep in mind is that if one dep is very popular, there's a higher chance it's already in people's deps tree. I'm not saying, we shouldn't replace a dependency if it's not popular, but it's something we should keep in mind too.

Anyway, after the lodash removal, installing svg-sprite leads to ~1700 files while in v2.0.0-beta5 it was ~2700 files.

strager:
Drop lodash.trim: #742

tommiv:
A note here that with the lodash.trim removal a small documentation bug is introduced:
```js
// Add SVG source files — the manual way ...
spriter.add('assets/svg-1.svg', null, fs.readFileSync('assets/svg-1.svg', 'utf-8'));
```
this line will throw an error because there's no null check on 
```js
name = trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
```

Possible solutions are 
1) to add a null-ish check
```js
name = trimStart(name ? name.trim() : '', `${path.sep}.`) || path.basename(file);
```

2) to force a string before .trim() call
```js
name = trimStart((name || '').trim(), `${path.sep}.`) || path.basename(file);
```

3) to update a readme
```
// Add SVG source files — the manual way ...
spriter.add('assets/svg-1.svg', '', fs.readFileSync('assets/svg-1.svg', 'utf-8'));
```

strager:
@tommiv Thanks for finding a bug in my patch!

@XhmikosR Please advise on the right solution.

jimCresswell:
I've opened #819 to remove lodash.trimstart, happy to make changes if that approach doesn't suit you or you find issues

baileys-li:
Since the next release will be a major release, this is a good opportunity to raise the minimum Node.js version to 22+ and use the built-in `glob` function instead of the library.

[...]

PR #749, opened 2022-11-02
refactor: drop `lodash.trim` dependency

XhmikosR:
Supersedes #742 

Refs #598

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -91,7 +91,7 @@ SVGSpriter.prototype.add = function(file, name, svg) {
             error = format('SVGSpriter.add: "%s" is not a valid relative file name', name);
         } else {
             name = trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
-            // TODO(strager): Avoid Buffer -> String -> Buffer conversion of svg.
+            // TODO: Avoid Buffer -> String -> Buffer conversion of svg.
             svg = svg.toString().trim();
 
             // Argument validation
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -14,7 +14,6 @@ const path = require('path');
 const events = require('events');
 const os = require('os');
 const { format } = require('util');
-const trim = require('lodash.trim');
 const trimStart = require('lodash.trimstart');
 const merge = require('lodash.merge');
 const File = require('vinyl');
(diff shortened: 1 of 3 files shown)


XhmikosR:
@Kreeg this seems to work fine from my tests. Does it LGTY?

lib/svg-sprite.js
```diff
@@ -84,15 +83,16 @@ SVGSpriter.prototype._namespacePow = [];
...
 
         // If the name part of the file path is absolute
         if (name && path.isAbsolute(name)) {
             error = format('SVGSpriter.add: "%s" is not a valid relative file name', name);
         } else {
-            name = trimStart(trim(name), `${path.sep}.`) || path.basename(file);
-            svg = trim(svg);
+            name = trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
+            // TODO: Avoid Buffer -> String -> Buffer conversion of svg.
+            svg = svg.toString().trim();
```
XhmikosR, line 95: Just wondering, would it make more sense to check if `svg` is Buffer with `Buffer.isBuffer()` and if so only then call `toString()`?

XhmikosR:
After a closer look, this is not identical to what we had before. Previously, if you passed `null` or `undefined`, it would proceed silently. Now we'll get a TypeError:

```
C:\Users\xmr\Desktop\svg-sprite\lib\svg-sprite.js:86
        file = file.trim();
                    ^

TypeError: Cannot read properties of undefined (reading 'trim')
    at SVGSpriter.add (C:\Users\xmr\Desktop\svg-sprite\lib\svg-sprite.js:86:21)
    at C:\Users\xmr\Desktop\svg-sprite\example.js:62:17
    at Array.forEach (<anonymous>)
    at addFixtureFiles (C:\Users\xmr\Desktop\svg-sprite\example.js:61:11)
    at Object.<anonymous> (C:\Users\xmr\Desktop\svg-sprite\example.js:71:1)
    at Module._compile (node:internal/modules/cjs/loader:1159:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1213:10)
    at Module.load (node:internal/modules/cjs/loader:1037:32)
    at Module._load (node:internal/modules/cjs/loader:878:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)

Node.js v18.12.1
```

lib/svg-sprite.js
Kreeg: This would seem a little overengineering for me.

Kreeg (approved):

PR #753, opened 2022-11-07
Update playwright-chromium

XhmikosR:


Kreeg:
@XhmikosR no, it isn't. The problem here is that i don't have much time to dig deeper. It's 99% that problem is in chromium rather than playwright.
I understand that we can't wait forever so let's land this. I will research sometime later.

XhmikosR:
If you work on battery, I wouldn't want to make things harder for you. We can still keep using the old version for more time.

Kreeg:
@XhmikosR apparently it's chromium bug. And it's caused without external display with mackbook m1 devices :/ Not with the power source as I though before. Ok let's keep the old version for now

PR #745, opened 2022-10-29
Fix `shape.id.separator` regression

sfe-efficy:
Hi!

I created a fork to fix the problem reported here: #740

Regards

Fixes #740

PR #755, opened 2022-11-08
lib/svg-sprite.js: revert change from e85b53c

### LINKED ISSUE #740: bug: 2.0.0: `shape.id.separator` does not produce an `id` with the parent directory names
In 1.5.0 `shape.id.separator` traversed full directory path + file name into the shape ID. Now in 2.0.0 shape ID is always a svg file name, without any parent folders.

XhmikosR:
It should really fix #740 

I'll try to release a new 2.0.2 version after this is merged.

EDIT: BTW this doesn't seem to be covered by the tests, any help adding a test case is welcome :)

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -125,7 +125,7 @@ SVGSpriter.prototype.add = function(file = '', name = '', svg = '') {
 
         // Instantiate a vinyl file
         file = new File({
-            base: path.dirname(file),
+            base: file.substring(0, file.length - name.length),
             path: file,
             contents: Buffer.from(svg)
         });
--- a/lib/svg-sprite/queue.js
+++ b/lib/svg-sprite/queue.js
@@ -39,7 +39,7 @@ module.exports = class SVGSpriterQueue extends EventEmitter {
      * @param {File} file                   Shape file
      */
     add(file) {
-        this._spriter.debug('Added "%s" to processing queue', path.basename(file.path));
+        this._spriter.debug('Added "%s" to processing queue', file.path.substr(file.base.length + path.sep.length));
         this._files.push(file);
         this.emit('add');
     }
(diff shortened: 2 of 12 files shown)


XhmikosR:
Note that we could probably revert #745 and go with the old code. Unsure if https://github.com/svg-sprite/svg-sprite/commit/e85b53c6343dd78d1a5eccae137e2b337be94a5a has caused more issues, so it might make more sense to revert it altogether.

XhmikosR:
Wouldn't it be better if we reverted both? See my comment above.

XhmikosR:
I mean, I'm not sure if we are missing something in tests and that patch has caused even more damage. I did it in a hurry and it might have caused more issues.

Kreeg:
This line is not covered yet by the unit tests. And all the snapshots are done with absolute svg paths. I'm not sure how to deal with this. We can just double up all the regression tests with the relative svgs

XhmikosR:
All I'm saying is that maybe it's safer for now if we just revert things to what it was and we can see if we can revisit this later. Since the old code did work.

Kreeg:
@XhmikosR do you want to revert both https://github.com/svg-sprite/svg-sprite/pull/745 and https://github.com/svg-sprite/svg-sprite/commit/e85b53c6343dd78d1a5eccae137e2b337be94a5a ?

XhmikosR:
Yeah, minus the tests from #745 if they still pass.

Kreeg:
Yes, let's do this

XhmikosR:
OK, so I tried it but there are issues in shape.js. Unsure how to proceed...

Kreeg:
I'll fix this tomorrow

Kreeg:
@XhmikosR please take a look. Previously all "files" in those test were fake mocks. Now they're real Vinyl files. I think this is correct

XhmikosR:
@Kreeg thanks! Does this mean that now tests will catch such regression in the future?

Is the PR ready from your side?

XhmikosR:
@Kreeg friendly ping :) Is this ready from your side? If so I'll merge it and cut a 2.0.2 later.

The next release should be a v3.0.0 assuming we manage to update svgo to 3.x and csso to 5.x :)

Kreeg:
@XhmikosR yes, I think it's ready

PR #759, opened 2022-11-12
Update codeql.yml

XhmikosR:


XhmikosR:
@Kreeg it seems there's a new CodeQL issue, but I'm not 100% convinced is correct. Feel free to tackle this later (it will show in the Security repo tab).

PR #760, opened 2022-11-12
Added file existence checking before adding to sprite.

Kreeg:
Found here #759
@XhmikosR I think those code scanning was correct. All the other file openings in bin file are wrapped in try...catch. So it's good to add another protection

### DIFF
--- a/bin/svg-sprite.js
+++ b/bin/svg-sprite.js
@@ -276,7 +276,17 @@ argv._.reduce((f, g) => [...f, ...glob.sync(g)], [])
             basename = basepos >= 0 ? basename.substr(basepos + 2) : path.basename(file);
         }
 
-        spriter.add(file, basename, fs.readFileSync(file));
+        let openedFile = null;
+
+        try {
+            openedFile = fs.readFileSync(file);
+        } catch (error) {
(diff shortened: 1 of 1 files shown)


Kreeg:
@XhmikosR sorry that was too early. I'll wrap it in try..catch instead

bin/svg-sprite.js
```diff
@@ -276,7 +276,17 @@ argv._.reduce((f, g) => [...f, ...glob.sync(g)], [])
             basename = basepos >= 0 ? basename.substr(basepos + 2) : path.basename(file);
         }
 
-        spriter.add(file, basename, fs.readFileSync(file));
+        let openedFile = null;
+
+        try {
+            openedFile = fs.readFileSync(file);
+        } catch (error) {
+            console.error(`[ERROR] Skipping (${file}) due to errors (${error.message.trim()})`);
```
XhmikosR, line 284: Wouldn't it be better to throw here?

It's hard working without tests for bin, but I assume you tested it anyway :)

Kreeg: I think you're right. It's better to throw. But github suggest using file descriptors instead. 
I think we can dismiss this alert

XhmikosR: Sounds good to me. We can revisit it later if needed.

PR #761, opened 2022-11-13
More refactoring

XhmikosR:
@Kreeg this should be the last batch of patches. The only patch I'm not 100% sure is the "Simplify a few boolean checks" one.

Check the non-whitespace diff: https://github.com/svg-sprite/svg-sprite/pull/761/files?w=1

TODO:

- [ ] Manually rebase patches if needed and don't squash when merging!

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -243,58 +243,60 @@ SVGSpriter.prototype.compileAsync = function(config) {
  * Run the next compile task
  */
 SVGSpriter.prototype._compile = function() {
-    // If the shape queue is not currently active
-    if (!this._queue.active && this._compileQueue.length) {
-        const [config, cb] = this._compileQueue.shift();
+    // If the shape queue is currently active return early
+    if (this._queue.active || this._compileQueue.length === 0) {
+        return;
+    }
 
-        // If this is a modeless run
-        if (Object.keys(config).length === 0) {
-            const files = {};
+    const [config, cb] = this._compileQueue.shift();
 
+    // If this is a modeless run
+    if (Object.keys(config).length === 0) {
+        const files = {};
+
+        // Add intermediate SVG files
+        if (this.config.shape.dest) {
+            files.shapes = this._getShapeFiles(this.config.shape.dest);
+            this.verbose('Returning %d intermediate SVG files', files.shapes.length);
+        }
+
+        this._logStats(files);
+        cb(null, files, {});
+    } else {
+        const masterShapes = this._shapes.filter(shape => !shape.master);
+        this.info('Compiling %d shapes...', masterShapes.length);
+
+        // Initialize the namespace powers
+        while (!this._namespacePow.length || (26 ** this._namespacePow.length < masterShapes.length)) {
+            this._namespacePow.unshift(26 ** this._namespacePow.length);
+
+            for (const shape of this._shapes) {
+                shape.resetNamespace();
+            }
+        }
+
+        // Sort shapes by ID
+        this._shapes = this._shapes.sort(this.config.shape.sort);
+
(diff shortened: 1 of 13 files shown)


lib/svg-sprite.js
```diff
@@ -323,13 +326,14 @@ SVGSpriter.prototype._indexNamespace = function(index) {
...
-    const tasks = [];
     const files = {};
     const layout = new Layouter(this, config);
     const createLayoutTask = (k, m) => _cb => {
         layout.layout(files, k, m, _cb);
     };
 
+    const tasks = [];
+
     for (const [mode, value] of Object.entries(config)) {
```
Kreeg, line 337: Maybe it's better to replace with `.map()` instead?

lib/svg-sprite/config.js
```diff
@@ -212,33 +212,38 @@ module.exports = class SVGSpriterConfig {
      * @private
      */
     _getMetaData() {
-        if (!('meta' in this.shape && !isPlainObject(this.shape.meta))) {
+        if (!('meta' in this.shape) || isPlainObject(this.shape.meta)) {
```
Kreeg, line 215: I think previous version was more obvious. New version can be shortened to  `if (isPlainObject(this.shape.meta)) because `isPlainObject(undefined) === false`

lib/svg-sprite/config.js
```diff
@@ -359,27 +369,36 @@ module.exports = class SVGSpriterConfig {
...
+
+            for (const [transformer, value] of Object.entries(transform)) {
+                if (value === true) {
+                    result.push([transformer, {}]);
+                    break;
+                }
+
+                if (isObject(value) || isFunction(value)) {
+                    result.push([transformer, value]);
+                    break;
```
Kreeg, line 401: There is no need to break here

XhmikosR:
@Kreeg feel free to push any changes, I think I'm done with the current pass of changes. I'm not 100% sure about your suggestions, but if you are confident about them, feel free to push them :)

Note that I can just partially revert the simplify boolean checks patch if you prefer the old way.

lib/svg-sprite/config.js
XhmikosR: You mean something like this?

```diff
 lib/svg-sprite/config.js | 8 +++-----
 1 file changed, 3 insertions(+), 5 deletions(-)

diff --git a/lib/svg-sprite/config.js b/lib/svg-sprite/config.js
index 4e32fe7..e0c59c0 100644
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -393,13 +393,11 @@ module.exports = class SVGSpriterConfig {
             for (const [transformer, value] of Object.entries(transform)) {
                 if (value === true) {
                     result.push([transformer, {}]);
-                    break;
-                }
-
-                if (isObject(value) || isFunction(value)) {
+                } else if (isObject(value) || isFunction(value)) {
                     result.push([transformer, value]);
-                    break;
                 }
+
+                break;
             }
         }
 
```

Kreeg: No, i mean the last `break` is redundant. There is no code left below this :)

XhmikosR: This is needed.

lib/svg-sprite.js
XhmikosR: Not sure if `map` would work, maybe we need `reduce`. I say we leave this for another time :)

XhmikosR:
@Kreeg can we proceed with this? Do you have any other suggestions?

Kreeg (approved):

Kreeg:
@XhmikosR seems OK

PR #768, opened 2022-12-06
Update sprite.js

XhmikosR:


XhmikosR:
I;m going to drop this, it doesn't look more readable, I think.

Kreeg:
@XhmikosR or we can add eslint-ignore?

PR #775, opened 2023-02-09
Fix `_transformShape` Not Calling All Transformations

### LINKED ISSUE #762: Bug: Spriter's Shape Transform Method Only Running First Transformation in shape.transform Array
Repo with minimum reproduction: https://github.com/babs20/svg-sprite-transform-bug

**Steps to Reproduce:**
1. Use the Spriter configuration below
2. Add SVG files to spriter instance
3. Run the compile method on the spriter
4. You will see that the output log only shows the optimization, but not the `console.log` in the second transformation. Also, if you inspect the output SVG you will see the custom transformations were not applied.

**Expected Behavior:**
SVGO optimizes shape, then my custom transformation runs on the shape.

**Actual Behavior:**
Whichever transformation comes first in the `shape.transform` array is run on the shape, but the subsequent transformations are not run.

**Sidenotes**
- This works with `svg-sprite@1.5.4`, but is broken in `svg-sprite@^2`
- I did a little digging into the source code and found that both transformations were being added to the `async.waterfall` correctly in `Spriter._transformShape` but, the second transformation function was not being called for some reason.
- I tried `node@12`, `node@14`, `node@16`, and `node@18`
- I am using Windows 11

Here is an example of my spriter config.
```js
const spriter = new SVGSpriter({
  dest: "",
  shape: {
    dimension: {
      maxWidth: 100,
      maxHeight: 100,
      precision: 2,
      attributes: false,
    },
    dest: "",
    transform: [
      "svgo",
      {
        custom: (shape, sprite, callback) => {
          console.log("2nd Transformation");
          const svgEl = shape.dom.getElementsByTagName("svg");
          const paths = shape.dom.getElementsByTagName("path");

          for (let i = 0; i < svgEl.length; i++) {
            svgEl[i].removeAttribute("fill");
          }

          for (let i = 0; i < paths.length; i++) {
            paths[i].setAttribute("fill", "");
          }

          callback(null);
        },
      },
    ],
  },
  expand: true,
  cwd: "sourceimages/sourceimages/svg",
  src: ["**/*.svg"],
  mode: {
    symbol: {
      bust: false,
      prefix: ".icon-%s",
      render: {
        scss: false,
        css: false,
      },
      example: true,
    },
  },
  log: "debug",
});

### PR DESCRIPTION
**Pull request by babs20**
Fixes #762 

This could also close #770

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -163,15 +163,15 @@ SVGSpriter.prototype._transformShape = function(shape, cb) {
     const createTransformTask = transform => {
         // If it's a custom transformer
         if (isFunction(transform[1])) {
-            return () => {
-                transform[1](shape, this, cb);
+            return callback => {
+                transform[1](shape, this, callback);
             };
         }
 
         // Else if it's a registered transformer
         if (transform[0] in this._shapeTransformers && isObject(transform[1])) {
-            return () => {
-                this._shapeTransformers[transform[0]](shape, transform[1], this, cb);
+            return callback => {
+                this._shapeTransformers[transform[0]](shape, transform[1], this, callback);
             };
         }
 
--- a/test/spriter.test.js
+++ b/test/spriter.test.js
@@ -112,3 +112,33 @@ describe('testing SVGSpriter', () => {
         });
     });
 });
+
+describe('testing transform', () => {
+    it('should call all transformations', async() => {
+        const testTransformFunction = jest.fn();
+        const spriter = new SVGSpriter({
+            shape: {
+                dest: 'svg',
+                transform: [
+                    'svgo',
+                    {
+                        custom(_, __, callback) {
+                            testTransformFunction();
+                            callback(null);
+                        }
+                    }
+                ]
+            }
+        });
+
+        spriter.add(new File({
+            base: path.dirname(TEST_SVG),
+            path: TEST_SVG,
+            contents: fs.readFileSync(path.join(__dirname, TEST_SVG))
+        }));
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/svg-sprite.js [hunk 1]
```diff
@@ -163,14 +163,14 @@ SVGSpriter.prototype._transformShape = function(shape, cb) {
     const createTransformTask = transform => {
         // If it's a custom transformer
         if (isFunction(transform[1])) {
-            return () => {
+            return cb => {
```
Kreeg, line 166: This name is shadowed by upper scope, pls change the name

lib/svg-sprite.js
```diff
@@ -163,14 +163,14 @@ SVGSpriter.prototype._transformShape = function(shape, cb) {
...
                 transform[1](shape, this, cb);
             };
         }
 
         // Else if it's a registered transformer
         if (transform[0] in this._shapeTransformers && isObject(transform[1])) {
-            return () => {
+            return cb => {
```
Kreeg, line 173: Here is the same

Kreeg (changes requested):
@babs20 hello! Thank you for contribution.

- Please fix minor issues here with naming
- Please squash all of your 3 commits into 1

Kreeg:
@babs20 hello! Please update commit message with more readable one, like `Fixed _transformShape not calling all transformations`

babs20:
@Kreeg Okay, I think I fixed everything 🥳 . Thanks for the feedback on the git flow 😄

Kreeg:
@babs20 thanks for help!

PR #797, opened 2023-04-12
Update devDependencies and fix new issues

XhmikosR:
@Kreeg: I haven't updated playwright-chromium due to that old issue you were facing. I assume you are still facing it with the latest version, right?

The only other deps that are outdated is glob and pngjs. For glob there's no hurry and in fact we better keep it to v7.x for now. pngjs I didn't try to update it, so if you have time feel free to update it :)

### DIFF
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -437,8 +437,8 @@ module.exports = class SVGSpriterConfig {
     /**
      * Prepare svg transforms
      *
-     * @param {object} svg         SVG object
-     * @returns {Array<Function>}  Array of svg transform functions
+     * @param {object} svg        SVG object
+     * @returns {Array|undefined} Array of svg transform functions
      * @private
      */
     _prepareAndProcessTransforms(svg) {
--- a/lib/svg-sprite/config.js
+++ b/lib/svg-sprite/config.js
@@ -324,10 +324,7 @@ module.exports = class SVGSpriterConfig {
         ['padding'].forEach(function(property) {
             let spacing;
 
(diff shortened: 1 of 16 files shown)


test/svg-sprite/config/config.constructor.shape.align.test.js
```diff
@@ -7,7 +7,7 @@ const yaml = require('js-yaml');
 const SVGSpriterConfig = require('../../../lib/svg-sprite/config.js');
 
 jest.mock('node:fs');
-jest.mock('yaml');
+jest.mock('js-yaml');
```
XhmikosR, line 10: I don't quite get how this worked all this time 😛

lib/svg-sprite/config.js
```diff
@@ -324,10 +324,7 @@ module.exports = class SVGSpriterConfig {
         ['padding'].forEach(function(property) {
             let spacing;
 
-            if (!Array.isArray(this.spacing[property])) {
```
Kreeg, line 327: Was this eslint's correction?

Kreeg:
Yes. The problem with playwright-chromium still exists. I do not have much time to deal with it and I sure this is a chromium problem rather than a playwright.
I updated pngjs. This was a seamless update, no extra work.

Kreeg (approved):

lib/svg-sprite/config.js
XhmikosR: Yeah, I guess the no negated condition rule I guess added some fixes

XhmikosR: Yeah, I guess the no negated condition rule I guess added some fixes

PR #798, opened 2023-04-12
Jest: limit maxWorkers, maxConcurrency and use workerThreads

XhmikosR:
The time difference hit isn't that big. We might even want to go with `os.cpus() - 4` or `os.cpus() / 2`.

Now Jest doesn't spawn 600 processes on my Windows VM.

CI seems to be slower, though. We could use a different number of threads for CI.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,8 +1,11 @@
 'use strict';
 
 const os = require('node:os');
+const process = require('node:process');
 
-const THREADS = Math.max(os.cpus().length - 2, 2);
+const THREADS = process.env.CI ?
+    os.cpus().length * 1.5 :
+    Math.max(os.cpus().length - 2, 2);
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


XhmikosR:
Did you try the branch? If it's not significantly slower, I'd like to keep it simple.

I don't see the point of spawning 600+ processes when running the tests, regardless of the OS.

Kreeg:
No I didn't. If it OK on your machine and CI - I'm OK with that.

Kreeg (approved):

XhmikosR:
Could you give it a go on your machine too? I'd like to see how big is the slowdown there before we merge this.

CI is the same, because it has 2 threads which is the minimum we set.

Kreeg:
@XhmikosR 34.548 s vs 30.104 s (config vs main). So it is OK.

PR #812, opened 2023-05-03
Switch to two-space indentation

XhmikosR:


Kreeg:
@XhmikosR can this be handled by xo? I noticed the only chande was made is for .editorconfig file

XhmikosR:
It's handled by xo, I changed the package.json too.

PR #815, opened 2023-05-09
tests: switch to async

XhmikosR:


test/helpers/add-files.js
```diff
@@ -11,13 +11,13 @@ const path = require('node:path');
  * @param {string} cwd                Working directory
  * @param {boolean} resolvePaths      Whether to resolve the paths of SVG files
  */
-function addFixtureFilesBase(spriter, files, cwd, resolvePaths) {
+async function addFixtureFilesBase(spriter, files, cwd, resolvePaths) {
   for (const file of files) {
```
XhmikosR, line 15: This is not optimal; I disabled the `no-await-in-loop` rule, but we should try to switch to `Promise.all` later.

Kreeg: this should be `Promise.all`. Otherwise there will be race conditions. 
Or switch to a new [statemets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of). I never tried it before though, but for me this is more preferable

XhmikosR:
I'll leave this for someone else to tackle later since it's not important. :)

PR #817, opened 2023-05-22
fix: fix crash with spriter.add(..., null, ...)

strager:
README.md has an example of calling spriter.add with name=null. Unfortunately, Git commit 23958e8c broke that case; calling spriter.add with name=null leads to an exception.

Make the code work like it did prior to commit 23958e8c: allow null, undefined, and empty strings.

Refs: 23958e8c9addb99036378e6a90e156f58797aacf

### DIFF
--- a/lib/svg-sprite.js
+++ b/lib/svg-sprite.js
@@ -92,7 +92,11 @@ SVGSpriter.prototype.add = function(file = '', name = '', svg = '') {
     if (name && path.isAbsolute(name)) {
       errorMessage = format('SVGSpriter.add: "%s" is not a valid relative file name', name);
     } else {
-      name = trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
+      if (name !== null) {
+        name = trimStart(name.trim(), `${path.sep}.`);
+      }
+
+      name = name || path.basename(file);
       // TODO: Avoid Buffer -> String -> Buffer conversion of svg.
       if (Buffer.isBuffer(svg)) {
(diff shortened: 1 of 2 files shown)


test/spriter.test.js
```diff
@@ -98,6 +98,23 @@ describe('testing SVGSpriter', () => {
         }));
       });
 
+      it.each([
+        [null],
+        [undefined],
+        ['']
+      ])('should create vinyl file with file name if name is %p', name => {
```
Kreeg, line 105: name here is always an array

Kreeg: it.each([
        null,
        undefined,
        ''
      ])
This is how it should be done

lib/svg-sprite.js
```diff
@@ -92,7 +92,11 @@ SVGSpriter.prototype.add = function(file = '', name = '', svg = '') {
     if (name && path.isAbsolute(name)) {
       errorMessage = format('SVGSpriter.add: "%s" is not a valid relative file name', name);
     } else {
-      name = trimStart(name.trim(), `${path.sep}.`) || path.basename(file);
+      if (name !== null) {
```
Kreeg, line 95: I would simply put `if (name)`

Kreeg (changes requested):
Hi there.
New test is false positive. Please update them and the corresponding code. Thanks

Kreeg:
Superseded by https://github.com/svg-sprite/svg-sprite/pull/819

PR #818, opened 2023-05-22
fix: improve error if svg.transforms is invalid

strager:
If the user provides {transform: 42}, svg.transform is reset to undefined. This leads to an error down the road such as "TypeError: this.transform is not iterable".

Explicitly error out earlier. This improves the message for users, and also fixes the following lint warning:

> JSDoc @returns declaration present but return expression not available
> in function.
> jsdoc/require-returns-check


Kreeg:
@strager hey! I added a corresponding test case. Pls merge my PR in your fork. Thanks

PR #865, opened 2023-11-22
Replace iterating over `Object.keys(NodeList)` with `NodeList.length`

Ingramz:
This works around xmldom versions greater than 0.8.8 & up to 0.8.10 seemingly not supporting use of Object.keys on NodeList to iterate over currently existing keys.

It should be noted I have not investigated whether `Object.keys(NodeList)` is allowed per DOM specification and it could be that there exists a regression in xmldom. Track https://github.com/xmldom/xmldom/issues/583 for relevant discussion.

Original patch from https://github.com/svg-sprite/svg-sprite/pull/835#issuecomment-1822484800

### DIFF
--- a/lib/svg-sprite/shape.js
+++ b/lib/svg-sprite/shape.js
@@ -247,7 +247,7 @@ SVGShape.prototype._stripInlineNamespaceDeclarations = function(element, nsMap) 
     }
   }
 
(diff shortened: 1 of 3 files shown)


XhmikosR:
@Kreeg LGTY? It's weird that this caused all the issues but it should be fine, at least up until the xmldom devs respond to the issue.

PR #905, opened 2024-05-10
SVGSpriter.add: clearer error message on no SVG

gurdiga:
## Summary

When I only see `SVGSpriter.add: You must provide SVG contents`, it hard for me to figure out what file is the culprit, so I’m adding the file path to the error message. :neckbeard:


XhmikosR:
This isn't right,  you changed the variable name.

Issue #942, opened 2025-08-11
`SVGSpriter.add: "..." is not the local part of "..."` error on Windows


jefferyto:
When source SVGs are in subdirectories I get this error on Windows 11 (PowerShell in regular Windows, not WSL):

```
PS C:\Users\Jeff\Downloads\svg-sprite-test> npm run compile:sprite

> svg-sprite-test@0.1.0 compile:sprite
> svg-sprite --log info --config sprite.config.json "images/./**/*.svg"

2025-08-11 14:08:05.050 - info: Using 4 threads
2025-08-11 14:08:05.055 - info: Created spriter instance
2025-08-11 14:08:05.110 - error: SVGSpriter.add: "icons/menu.svg" is not the local part of "C:\Users\Jeff\Downloads\svg-sprite-test\images\icons\menu.svg" SVGSpriter.add: "icons/menu.svg" is not the local part of "C:\Users\Jeff\Downloads\svg-sprite-test\images\icons\menu.svg"
C:\Users\Jeff\Downloads\svg-sprite-test\node_modules\svg-sprite\lib\svg-sprite.js:125
            throw e;
            ^

ArgumentError: SVGSpriter.add: "icons/menu.svg" is not the local part of "C:\Users\Jeff\Downloads\svg-sprite-test\images\icons\menu.svg"
    at SVGSpriter.add (C:\Users\Jeff\Downloads\svg-sprite-test\node_modules\svg-sprite\lib\svg-sprite.js:123:23)
    at C:\Users\Jeff\Downloads\svg-sprite-test\node_modules\svg-sprite\bin\svg-sprite.js:279:17
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (C:\Users\Jeff\Downloads\svg-sprite-test\node_modules\svg-sprite\bin\svg-sprite.js:267:6)
    at Module._compile (node:internal/modules/cjs/loader:1688:14)
    at Object..js (node:internal/modules/cjs/loader:1820:10)
    at Module.load (node:internal/modules/cjs/loader:1423:32)
    at Function._load (node:internal/modules/cjs/loader:1246:12)
    at TracingChannel.traceSync (node:diagnostics_channel:322:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:235:24)

Node.js v22.18.0
PS C:\Users\Jeff\Downloads\svg-sprite-test>
```

package.json:

```json
{
  "name": "svg-sprite-test",
  "version": "0.1.0",
[...]

PR #944, opened 2025-08-12
svg-sprite.js: Normalize `file` and `name` paths

jefferyto:
This uses `path.normalize()` to ensure `file` and `name` values use the same path separators.

This also tests if both variables are not the empty string before calling `path.normalize()` as it returns `.` in that case.

`path.normalize()` also removes `./` segments so it is not necessary to call `trimStart()` on `name`.

Fixes https://github.com/svg-sprite/svg-sprite/issues/942


jefferyto:
The lint errors have been fixed.

PR #921, opened 2024-09-03
refactor: replace 'glob' package with 'fdir'

ZuBB:
In my project, I am getting the next warning

> npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported

I decide to step in and give it a try. All seems to work good.

As a side note documentation still references `glog` but only in examples as an external (read brought by end-user) dependency. I think its Ok to have it in that role

Refs https://github.com/svg-sprite/svg-sprite/issues/598, https://github.com/svg-sprite/svg-sprite/issues/889.
Discards https://github.com/svg-sprite/svg-sprite/pull/916

Issue #889, opened 2024-03-05
Glob - Inflight - Missing Release of Resource after Effective Lifetime


Luen:
Security vulnerability identified by Snyk within the `svg-sprite` package's dependencies. 
The issue is tied to the `inflight` package, which is a transitive dependency through `glob`. According to Snyk, the vulnerability is registered as [SNYK-JS-INFLIGHT-6095116](https://security.snyk.io/vuln/SNYK-JS-INFLIGHT-6095116) and [CWE-772: Missing Release of Resource after Effective Lifetime](https://cwe.mitre.org/data/definitions/772.html).

### Issue Description
The `inflight` package, which is used to prevent parallel execution of async tasks with the same key, has been reported to have a security vulnerability. 

### Relevant Context
`glob` version 9 and onwards have moved away from using callbacks to promises, leading to the removal of `inflight` from its dependencies due to the API changes. This evolution is discussed in the following `node-glob` GitHub issues:
- [isaacs/node-glob#435](https://github.com/isaacs/node-glob/issues/435)
- [isaacs/node-glob#573](https://github.com/isaacs/node-glob/issues/573)

Additionally, the `inflight` package itself has it's own issues:
- [isaacs/inflight#5](https://github.com/isaacs/inflight/issues/5)
- [isaacs/inflight#19](https://github.com/isaacs/inflight/issues/19)

### Suggested Action
Given that `glob` has evolved past the need for `inflight`, it may be beneficial for `svg-sprite` to update its dependencies accordingly to a version of `glob` that does not rely on `inflight` V9/V10.

PR #975, opened 2026-05-02
refactor: replace `glob` with Node.js API

baileys-li:
Same as from #921, #973, #974, but instead of replacing `glob` with another library use Node.js native API.
Refs #598, #889

Pros:
* Fewer dependencies; it uses `node:fs`, which was already in use
* Since it’s a built-in API, it’s more future-proof
* Runs faster; on my machine, Jest tests took 22.454 seconds. I think this isn’t so much because the API itself is faster, but rather because the Node.js version has been updated.

Cons:
* **Breaking change**. The minimal Node version has been raised, since the glob API was introduced in Node 22+. But version 22 was released two years ago. Everyone who could update has already done so.


joeda1:
@baileys-li I did a fork of this package and add your change because it looks like the package unmaintained. It seems to work well. Thanks. https://github.com/svgforge

joeda1:
@baileys-li I found now a issue this change causing.

According the docs you can use the following syntax:
```
svg-sprite -C material.json  'material/./**/*.svg'
```

Then `material` does not become a part of the id mapping. This is really handy feature that is broken with a switch to `fs.glob`
I found a workaround in my fork and applied it already: https://github.com/svgforge/svgforge-cli/blob/main/bin/svgforge.js#L529

@XhmikosR @Kreeg Are you still around and manage the project?

